PR #24, opened 2019-09-04
Fix up npm security warnings

roborourke:
🔥 lodash


roborourke:
Yeah why not. I think the `npm` stuff isn't that likely to break in this repo.

PR #26, opened 2019-09-09
Load analytics.js async

### LINKED ISSUE #25: Performance: Caching and size reduction for analytics.js
We need to trim down the initial analytics.js bundle because it's a blocking script and loaded very early.

We ideally need to keep the early loading as the experiments plugin relies on it and we want our custom components to render immediately.

Possible approaches could involve one or more of the following:

- Rewrite everything to use vanilla JS and zero (or minimal) dependencies
- Lambda@Edge function to inline script
  - Follow strategy here: https://www.filamentgroup.com/lab/inlining-cache.html
  - Service worker cache inlined file
  - Set cookie
  - Bypass in lambda if cookie set
- Always inline the script 😱 ~40kb added to pages
- Use async and provide an alternative way to load/queue dependencies (maybe https://github.com/filamentgroup/loadJS)

roborourke:
This does two things and improves the Lighthouse report as regards blocking scripts:

- Adds `async` attribute to the analytics.js script tag
- Moves the `registerAttribute` and `registerMetric` functions to the inline script to support experiments.js

Fixes #25

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -11,6 +11,8 @@ namespace Altis\Analytics;
 require_once ROOT_DIR . '/inc/utils.php';
 
 function setup() {
+	// Handle async scripts.
+	add_filter( 'script_loader_tag', __NAMESPACE__ . '\\async_scripts', 20, 2 );
 	// Load analytics scripts super early.
 	add_action( 'wp_head', __NAMESPACE__ . '\\enqueue_scripts', 0 );
 }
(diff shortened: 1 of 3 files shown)


roborourke:
I think it's still worth trying to trim down the analytics.js script as it's twice the size of Google's but this provides a pretty significant improvement already.

inc/namespace.php
```diff
@@ -130,15 +132,39 @@ function get_client_side_data() : array {
...
+ * @return string The script tag markup.
+ */
+function async_scripts( string $tag, string $handle ) : string {
+	global $wp_scripts;
+
+	if ( ! $wp_scripts->get_data( $handle, 'async' ) || strpos( $tag, 'async' ) !== false ) {
+		return $tag;
+	}
+
+	$tag = str_replace( '></script>', ' async></script>', $tag );
```
joehoyle: Sheesh, does wp_enqueue_script not support this?

roborourke: Sure doesn't!

roborourke:
@rmccue / @joehoyle anything to change here?

inc/namespace.php
```diff
@@ -130,15 +132,39 @@ function get_client_side_data() : array {
...
+	global $wp_scripts;
+
 	wp_enqueue_script( 'altis-analytics', plugins_url( 'build/analytics.js', __DIR__ ), [], null, false );
 	wp_add_inline_script(
 		'altis-analytics',
 		sprintf(
-			'var Altis = Altis || {}; Altis.Analytics = %s;', wp_json_encode(
+			'var Altis = Altis || {}; Altis.Analytics = %s;' .
+			'Altis.Analytics.registerAttribute = function (key, value) { Altis.Analytics._attributes[key] = value; };' .
+			'Altis.Analytics.registerMetric = function (key, value) { Altis.Analytics._metrics[key] = value; };',
```
rmccue: This is getting to be a little complex now; I wonder whether we could separate this out a bit 🤔

roborourke: I think we can refactor this whole script down some in a future issue to reduce the size of it. Functionally however this is no different than what was there previously, just these functions need to available super early.

PR #32, opened 2020-02-13
Disable data logging on preview pages.

onyekaa:
We don't want to log  Preview pages so I send an empty data object and check for this when calling the record function.


src/analytics.js
```diff
@@ -436,12 +436,14 @@ window.addEventListener("beforeunload", async () => {
...
-window.Altis.Analytics.updateEndpoint = Analytics.updateEndpoint;
-window.Altis.Analytics.record = (type, data = {}) =>
-	Analytics.record(
-		type,
-		{
-			attributes: getAttributes(data.attributes || {}),
-			metrics: getMetrics(data.metrics || {})
-		}
-	);
+if (!Object.keys(Data).length === 0) {
```
roborourke: The problem with this approach is that these functions will be undefined during a preview, however other code may still be trying to use them.

The goal here is to make these functions into no-op versions that do nothing eg. `window.Altis.Analytics.updateEndpoint = () => {}`

inc/namespace.php
```diff
@@ -29,7 +29,10 @@ function setup() {
  * @return array
  */
 function get_client_side_data() : array {
-
+	// Return empty object if preview.
+	if ( is_preview() ) {
```
roborourke: We may still need the data for other things like the AB test components. I'll add my suggestion for the approach in the main review comment.

src/analytics.js
onyekaa: Oh I figured they were using other parts of the file, not the actual data. Will look at this again.

roborourke: To clarify other code doesn't have access to anything in this file - just this bit at the end, we're extending the `window` object to expose these functions. Everything else is contained.

onyekaa:
I've updated the PR to use a filter to check for the post's preview status, and disable the update/record functions if true.

src/analytics.js
```diff
@@ -435,13 +436,18 @@ window.addEventListener("beforeunload", async () => {
...
-window.Altis.Analytics.record = (type, data = {}) =>
-	Analytics.record(
-		type,
-		{
-			attributes: getAttributes(data.attributes || {}),
-			metrics: getMetrics(data.metrics || {})
-		}
-	);
+// Expose userland API. Return empty functions if the Noop field is true.
+if (Noop) {
```
roborourke: I think this will create the same problems as before because we're only exposing the functions on the `window` object when this is false - that means something else could still try and call `window.Altis.Analytics.record()` but the function won't exist.

You would have seen this stop the default built in events from firing because it's updating the functions that they would be using but the last part of the testing should ensure that you can call `Altis.Analytics.record( 'event' )` from the console and nothing happening - but also no error thrown.

You could remove the `} else {` part so that the window object always gets updated, either with the "live" version of the functions or "no-op" version.

src/analytics.js
```diff
@@ -435,13 +436,18 @@ window.addEventListener("beforeunload", async () => {
...
+	window.Altis.Analytics.updateEndpoint = Analytics.updateEndpoint;
+	window.Altis.Analytics.record = (type, data = {}) =>
+		Analytics.record(
+			type,
+			{
+				attributes: getAttributes(data.attributes || {}),
+				metrics: getMetrics(data.metrics || {})
+			}
+		);
+}
```
roborourke: Make sure to configure your editor to add a blank line at the end of your files, it's a more of a habit from writing shell scripts so a program know when to terminate reading a file but it's also part of our coding standards.

src/analytics.js
```diff
@@ -16,6 +16,7 @@ const {
 	_metrics,
 	Config,
 	Data,
+	Noop,
 } = Altis.Analytics;
```
roborourke: I was thinking here you could check the `Noop` value early, then set the empty functions and `return` preventing further execution of the file.

```suggestion
	Noop,
} = Altis.Analytics;

// Expose no-op functions to browser for specific contexts. 
if (Noop) {
	window.Altis.Analytics.updateEndpoint = () => {};
	window.Altis.Analytics.record = () => {};
	return;
}
```

inc/namespace.php
```diff
@@ -176,6 +185,7 @@ function enqueue_scripts() {
 						'CognitoRegion' => defined( 'ALTIS_ANALYTICS_COGNITO_REGION' ) ? ALTIS_ANALYTICS_COGNITO_REGION : null,
 						'CognitoEndpoint' => defined( 'ALTIS_ANALYTICS_COGNITO_ENDPOINT' ) ? ALTIS_ANALYTICS_COGNITO_ENDPOINT : null,
 					],
+					'Noop' => (bool) apply_filters( 'altis.analytics.preview', false ),
```
roborourke: Lets give this filter a more generic name - the preview is just one context we've found where we don't want to record events so let's call it `altis.analytics.noop`.

We should document the filter on the readme too.

roborourke: Also it's good practice to add docblocks for filters and actions. The cleanest way to do this would be to set the variable on a line above the `wp_enqueue_script()` call like so:

```php
/**
 * The no-op filter allows you to prevent any event tracking or user endpoint
 * updates without needing to update any other code. This is used when previewing
 * posts for example.
 *
 * @param bool $noop If true prevents any analytics events being recorded.
 */
$noop = (bool) apply_filters( 'altis.analytics.noop', false );
```

src/analytics.js
roborourke: Whoops - I made a mistake here, in JS you can't return outside of a function, but simple solution is to wrap the entire thing after the `import`s in a closure function eg:

```js
import ... etc...

( function () {
   // the whole file contents after import goes here
} )();
```

PR #33, opened 2020-02-14
Schedule deletion of index created a week in the past.

onyekaa:
What I did here was make a DELETE call to an index named one week in the past. The assumptions here are that the index is rotated daily which appears to be the default. My next concern here is what happens if for some reason the deletion fails. 

Ideally I'd like to find all indexes with names older than a certain date, but I'm not sure if that's the right approach because it requires fetching all indexes, then checking the names one by one, extracting the date format and comparing against the current time. If this makes more sense I can do that.


inc/utils.php
```diff
@@ -122,6 +123,63 @@ function query( array $query, array $params = [] ) : ?array {
...
+	if ( wp_remote_retrieve_response_code( $response ) !== 200 || is_wp_error( $response ) ) {
+		if ( is_wp_error( $response ) ) {
+			trigger_error( sprintf(
+				"Analytics: ES index deletion failed: %s",
+				$response->get_error_message()
+			), E_USER_WARNING );
+		} else {
+			trigger_error( sprintf(
+				"Analytics: ES index deletion failed:\n%s\n%s",
+				json_encode( $index ),
```
roborourke: I think it'll be easier to read the logs if this uses `JSON_PRETTY_PRINT` and a few other useful flags for output eg:
```suggestion
				json_encode( $index, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE ),
```

inc/utils.php
```diff
@@ -122,6 +123,63 @@ function query( array $query, array $params = [] ) : ?array {
...
 }
 
+/**
+ * Delete old Analytics indexes in Elasticsearch.
+ *
+ * @return array|null
+ */
+function delete_old_indexes() : ?array {
+	// Index age in days.
+	$duration = get_config()['modules']['analytics']['max_index_age'] ?? 7;
```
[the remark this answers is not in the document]
onyekaa: Changed this to use a filter, which I've tested with altis-analytics.

inc/utils.php
```diff
@@ -122,6 +123,63 @@ function query( array $query, array $params = [] ) : ?array {
...
+
+	return $result;
+}
+
+add_action( 'index_maintenance', 'delete_old_indexes' );
+
+// Schedule index maintenance daily.
+if( !wp_next_scheduled( __NAMESPACE__ . '\\index_maintenance' ) ) {
+	wp_schedule_event( time(), 'daily',  __NAMESPACE__ . '\\index_maintenance' );
+}
```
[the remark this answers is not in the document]
onyekaa: Ended up moving everything to namespace.

inc/utils.php
```diff
@@ -122,6 +123,63 @@ function query( array $query, array $params = [] ) : ?array {
...
+
+	if ( json_last_error() ) {
+		trigger_error( 'Analytics: ES deletion response could not be decoded.', E_USER_WARNING );
+		return null;
+	}
+
+	return $result;
+}
+
+add_action( 'index_maintenance', 'delete_old_indexes' );
```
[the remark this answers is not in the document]
onyekaa: I think I actually added this for testing, hence the weird location. This has been fixed.

inc/namespace.php
```diff
@@ -196,3 +198,77 @@ function enqueue_scripts() {
 	// Print queued scripts.
 	print_head_scripts();
 }
+
+/**
+ * Schedule index maintenance daily.
+ */
+if( ! wp_next_scheduled( __NAMESPACE__ . '\\index_maintenance' ) ) {
+	wp_schedule_event( time(), 'daily',  __NAMESPACE__ . '\\index_maintenance' );
+}
```
roborourke: Again - you should avoid calling functions directly in a file, what I mean by mixing side effects and declaring functions is something like the following:

```php
// declaring a function
function foo() {
    return 'foo';
}
// calling a function
bar();
```

In this case you should move this block of code inside the `setup()` function.

inc/namespace.php
```diff
@@ -196,3 +198,77 @@ function enqueue_scripts() {
...
+ */
+function delete_old_indexes() : ?array {
+	// Index age in days.
+	$duration = apply_filters( 'altis.analytics.index_age', 7 );
+
+	// If duration has been set out of the 60-day limit, default to 7 days and warn user.
+	if ( $duration < 1 || $duration > 60 ) {
+		$duration = 7;
+		trigger_error( 
+			'Analytics data retention period must be between 1 and 60 days, defaulting to 7.',
```
roborourke: Rather than completely change the number of days we can move the retention period to the nearest value instead using `min()` and `max()`:
```suggestion
		$duration = max( 1, min( $duration, 60 ) );
		trigger_error( 
			sprintf( 'Analytics data retention period must be between 1 and 60 days. This has been set to %d.', $duration ), 
```

inc/namespace.php
```diff
@@ -196,3 +198,77 @@ function enqueue_scripts() {
...
+	wp_schedule_event( time(), 'daily',  __NAMESPACE__ . '\\index_maintenance' );
+}
+
+
+/**
+ * Delete old Analytics indexes in Elasticsearch.
+ */
+function delete_old_indexes() : ?array {
+	// Index age in days.
+	$duration = apply_filters( 'altis.analytics.index_age', 7 );
```
roborourke: Lets cast this to an integer, and we can make the filter name a bit more descriptive too:
```suggestion
	$duration = (int) apply_filters( 'altis.analytics.max_index_age', 7 );
```

inc/namespace.php
```diff
@@ -196,3 +198,77 @@ function enqueue_scripts() {
...
+if( ! wp_next_scheduled( __NAMESPACE__ . '\\index_maintenance' ) ) {
+	wp_schedule_event( time(), 'daily',  __NAMESPACE__ . '\\index_maintenance' );
+}
+
+
+/**
+ * Delete old Analytics indexes in Elasticsearch.
+ */
+function delete_old_indexes() : ?array {
+	// Index age in days.
```
roborourke: For filters we can document them like we would a function, so a docblock with params and an explanation. This will help us in future as we can generate documentation from these:
```suggestion
	/**
	 * Filter the maximum index age in days.
	 *
	 * @param int $days The maximum number of days to keep indexes for, defaults to 7.
	 */
```

inc/namespace.php
```diff
@@ -196,3 +198,77 @@ function enqueue_scripts() {
...
+ */
+if( ! wp_next_scheduled( __NAMESPACE__ . '\\index_maintenance' ) ) {
+	wp_schedule_event( time(), 'daily',  __NAMESPACE__ . '\\index_maintenance' );
+}
+
+
+/**
+ * Delete old Analytics indexes in Elasticsearch.
+ */
+function delete_old_indexes() : ?array {
```
roborourke: This function should probably return `true` when successful and a `WP_Error` if it fails thinking about it. As it stands it will always return `null` - there's no return value at the end of the function and when it has errors it explicitly returns `null`.

inc/namespace.php
```diff
@@ -196,3 +198,77 @@ function enqueue_scripts() {
...
+		);
+	}
+
+	// Get index name by date. 
+	$date = new \DateTime();
+	$date->sub( new \DateInterval( 'P'. $duration . 'D' ) );
+	$index_date = ( $date->format('U') );
+
+	// Get URL.
+	$index_list = wp_remote_request( get_elasticsearch_url() . '/analytics-*/');
```
roborourke: Keep an eye on your code style here. It's small stuff but makes a big difference to readability - I think it'll be worth looking at getting phpcs set up properly in vscode with the HM Coding Standards then it'll catch this stuff as you go
```suggestion
	$index_list = wp_remote_request( get_elasticsearch_url() . '/analytics-*/' );
```

inc/namespace.php
```diff
@@ -196,3 +198,77 @@ function enqueue_scripts() {
...
+
+	$indices = json_decode( wp_remote_retrieve_body( $index_list ), true);
```
roborourke: Need to check whether the response was successful here first.
```suggestion
	$indices = json_decode( wp_remote_retrieve_body( $index_list ), true );
```

inc/namespace.php
```diff
@@ -196,3 +198,77 @@ function enqueue_scripts() {
...
+	
+	// Check each available index's creation date, and delete if too old. 
+	foreach ( $indices as $i ) {
+		if ( $i['settings']['index']['creation_date'] < $index_date ) {
```
roborourke: There's a good habit to get into to avoid writing large indented `if` blocks that make it tricky to track the logic in the code which is to reverse the logic and return early, or in the case of a loop to `continue` early eg:
```suggestion
		if ( $i['settings']['index']['creation_date'] >= $index_date ) {
			continue;
		}
```

inc/namespace.php
```diff
@@ -196,3 +198,77 @@ function enqueue_scripts() {
...
+				$json = wp_remote_retrieve_body( $response );
+				$result = json_decode( $json, true );
+			
+				if ( json_last_error() ) {
+					trigger_error( 'Analytics: ES deletion response could not be decoded.', E_USER_WARNING );
+					return null;
+				}
+			}
+		}
+	}
```
roborourke: Lets return `true` here if we get this far to indicate success.

inc/namespace.php
```diff
@@ -196,3 +198,77 @@ function enqueue_scripts() {
...
+}
```
roborourke: Newline at end of file
```suggestion
}

```

inc/namespace.php
```diff
@@ -196,3 +198,77 @@ function enqueue_scripts() {
...
+	$indices = json_decode( wp_remote_retrieve_body( $index_list ), true);
+	
+	// Check each available index's creation date, and delete if too old. 
+	foreach ( $indices as $i ) {
+		if ( $i['settings']['index']['creation_date'] < $index_date ) {
+			// Get the name of the index to delete;
+			$index = $i['settings']['index']['provided_name'];
+
+			// Create new deletion URL.
+			$delete_url = add_query_arg( [], get_elasticsearch_url() . '/' . $index . '/' );
```
roborourke: Doesn't look like `add_query_arg()` is doing anything here, guessing it's from something you were testing out.

inc/namespace.php
```diff
@@ -196,3 +198,77 @@ function enqueue_scripts() {
...
+	if ( $duration < 1 || $duration > 60 ) {
+		$duration = 7;
+		trigger_error( 
+			'Analytics data retention period must be between 1 and 60 days, defaulting to 7.', 
+			E_USER_WARNING 
+		);
+	}
+
+	// Get index name by date. 
+	$date = new \DateTime();
```
roborourke: Rather than `\DateTime()` it helps PHP optimise itself a bit to add `use DateTime;` at the top of the file then you don't need the backslash here. Not a blocker really but it's consistent with our coding standards.

inc/namespace.php
```diff
@@ -196,3 +198,77 @@ function enqueue_scripts() {
 	// Print queued scripts.
 	print_head_scripts();
 }
+
+/**
+ * Schedule index maintenance daily.
+ */
+if( ! wp_next_scheduled( __NAMESPACE__ . '\\index_maintenance' ) ) {
+	wp_schedule_event( time(), 'daily',  __NAMESPACE__ . '\\index_maintenance' );
```
roborourke: Do think it'd make sense to get the nearest time at midnight? So indexes are only removed at the end of a day rather than whatever time this code runs initially?

PR #35, opened 2020-02-27
Add script tag version string

roborourke:
Should avoid issues with browser security integrity checks.

### DIFF
--- a/.circleci/deploy.sh
+++ b/.circleci/deploy.sh
@@ -62,6 +62,10 @@ fi
 
 rsync -av "$SRC_DIR/" "$BUILD_DIR" --exclude-from "$SRC_DIR/.circleci/deploy-exclude.txt"
 
+# Swap commit placeholders
+find . -name '*.php' | xargs sed -i .bak 's/__COMMIT__/'"$COMMIT"'/g'
+find . -name '*.bak' -delete
+
 # Add changed files
 git add .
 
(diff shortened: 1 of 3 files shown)


.circleci/deploy.sh
```diff
@@ -62,6 +62,10 @@ fi
 
 rsync -av "$SRC_DIR/" "$BUILD_DIR" --exclude-from "$SRC_DIR/.circleci/deploy-exclude.txt"
 
+# Swap commit placeholders
+find . -name '*.php' | xargs sed -i .bak 's/__COMMIT__/'"$COMMIT"'/g'
```
rmccue: This'll bust the cache on _every_ commit; we could read the version string from plugin.php instead?

roborourke: The plugin version string doesn't necessarily reflect if there were changes to `analytics.js` either, I was thinking this way we know for sure that the cache will be busted. But it's only if the plugin version is updated that a build & release happens anyway.

Hm, maybe it's best to leave this and just be really careful about ensuring there are updates to the version string if the JS has changed.

rmccue: We could have a content-based hash? `HASH=(md5 < analytics.js)`

roborourke: Thought that was a bit heavy to do on every run but could do, ah - you mean in `deploy.sh` 🤔

PR #38, opened 2020-03-12
Improve endpoint updates

roborourke:
It is possible to just use the put events command to update endpoints so we can remove the update endpoint command requirement. This also has the effect of ensuring the endpoint is always in sync with the event records as there is no race condition between the initial update endpoint call and the first event being recorded.

Lastly it also removes an unnecessary HTTP request to update endpoint separately.


roborourke:
I'll do a full code-style sweep in a separate PR 👍

PR #39, opened 2020-03-13
Remove custom eslintrc

rmccue:
This should force hmlinter to fall back to the regular HM coding
standards.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,3 +1,0 @@
-{
-	"parser": "babel-eslint"
-}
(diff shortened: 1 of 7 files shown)


.eslintrc
```diff
@@ -1,3 +1,10 @@
 {
-	"parser": "babel-eslint"
+	"extends": "humanmade",
+	"globals": {
+		"Altis": "readonly"
```
rmccue: Do we need this? I think using `window.Altis` instead avoids the need to change this?

PR #40, opened 2020-03-13
WIP: Use navigator.sendBeacon on window.unload

roborourke:
This change ensures that no analytics data is lost and will improve the results of queries for current active users and time on page.

In addition it will enhance performance as it removes the blocking ajax request on beforeunload for browsers that support it.

### DIFF
--- a/inc/beacon/namespace.php
+++ b/inc/beacon/namespace.php
@@ -1,0 +1,160 @@
+<?php
+/**
+ * Beacon API endpoint for loggin events on window.unload.
+ *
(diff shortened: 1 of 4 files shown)


roborourke:
Hmm, thinking about this a bit more it is not a good idea to do a POST on every page view to the web servers! Maybe needs to be a lambda instead or we try and speak to someone at AWS about getting support for `sendBeacon` in their JS SDK.

roborourke:
Closing this out, while it works it's a non starter for performance reasons

PR #42, opened 2020-03-16
Add IE11 support

### LINKED ISSUE #37: Error in IE11: URLSearchParams is undefined
One of our Altis sites requires IE11 support, and this module utilizes the [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) interface which is not available in that environment.

It looks like there's at least [one extant polyfill](https://www.npmjs.com/package/url-search-params-polyfill), though I haven't vetted it yet. Alternatively we could accept that analytics will not work properly in IE11 because it won't have _that_ much marketshare, but it shouldn't be too onerous to fix properly.

roborourke:
Fixes #37

This also simplifies and removes duplication of some of the complex es6 syntax around building the attributes objects to send with events.

### DIFF
--- a/src/analytics.js
+++ b/src/analytics.js
@@ -5,11 +5,11 @@ import UAParser from 'ua-parser-js';
 import merge from 'deepmerge';
 
 // AWS SDK.
-import { CognitoIdentityClient } from "@aws-sdk/client-cognito-identity-browser/CognitoIdentityClient";
-import { GetCredentialsForIdentityCommand } from "@aws-sdk/client-cognito-identity-browser/commands/GetCredentialsForIdentityCommand";
-import { GetIdCommand } from "@aws-sdk/client-cognito-identity-browser/commands/GetIdCommand";
-import { PinpointClient } from "@aws-sdk/client-pinpoint-browser/PinpointClient";
-import { PutEventsCommand } from "@aws-sdk/client-pinpoint-browser/commands/PutEventsCommand";
+import { CognitoIdentityClient } from '@aws-sdk/client-cognito-identity-browser/CognitoIdentityClient';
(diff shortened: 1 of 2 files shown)


roborourke:
@kadamwhite bumping this, any chance of a review?

roborourke:
Will wait for #47 to go in first as there's some overlap regarding the `getAttributes()` and `getMetrics()` functions.

roborourke:
Updated to handle all query vars the same way, they were needlessly duplicated before.

src/analytics.js
```diff
@@ -62,16 +69,13 @@ window.addEventListener( 'scroll', () => {
...
-const params = new URLSearchParams( window.location.search );
-const utm = {
-	utm_source: params.get( 'utm_source' ) || '',
-	utm_medium: params.get( 'utm_medium' ) || '',
-	utm_campaign: params.get( 'utm_campaign' ) || '',
-	utm_term: params.get( 'utm_term' ) || '',
-	utm_content: params.get( 'utm_content' ) || '',
-};
+const params = parseQueryString( window.location.search );
+const qv_params = {};
```
kadamwhite, line 75: nitpick: `qvParams`?

src/utils/index.js
```diff
@@ -32,3 +32,59 @@ export const throttle = ( delay, fn ) => {
...
+	}
+	if ( ! Array.isArray( value ) ) {
+		value = [ value ];
+	}
+	return value.map( sanitizeCallback );
+};
+
+const sanitizeAttribute = value => value.toString();
+
+const sanitizeMetric = value => parseFloat( Number( value ) );
```
kadamwhite, line 62: It's pretty obvious what these do but docbloc presence is inconsistent in this file

src/analytics.js
roborourke: Yep, don't know I had a blank on that one!

roborourke:
Thanks @kadamwhite, much appreciated

PR #36, opened 2020-02-28
Audience UI

roborourke:
Still plenty to do and questions to answer. Like should we grab the index mapping to access all the fields so we get their type, or just infer that given we know the data structure...

Fixes #19 

TODO:

- [x] Audience test endpoint
- [x] Audience save endpoint
- [x] Handle lack of analytics data

Potentially split these to separate PRs

- [ ] Documentation
- [x] Handle numbers via ranges eg. gt, lt, gte, lte
- [x] Handle wildcards / free text
- [x] Audience post listing columns - estimate spark line / number
- [ ] Audience add/edit control & modal for block editor
- [ ] Assign endpoint to audience client side
- [ ] Provide on updateEndpoint event for external code to hook in and re-render things

PR #43, opened 2020-04-02
Iterate on PHP for audience UI

rmccue:
Fixes a bunch of the smaller review comments in #36. These are the non-controversial parts. :)

### DIFF
--- a/inc/audiences/namespace.php
+++ b/inc/audiences/namespace.php
@@ -100,7 +100,7 @@ function audience_ui( WP_Post $post ) {
 		return;
 	}
 
-	echo sprintf(
+	printf(
 		'<div id="altis-analytics-audience-ui" data-post-id="%d" data-audience="%s" data-fields="%s">' .
 		'<p class="loading"><span class="spinner is-active"></span> %s</p>' .
 		'<noscript><div class="error msg">%s</div></noscript>' .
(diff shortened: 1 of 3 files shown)


inc/audiences/namespace.php
```diff
@@ -325,29 +324,51 @@ function get_estimate( array $audience ) : ?array {
...
 			'histogram' => [
 				'histogram' => [
 					'field' => 'event_timestamp',
-					'interval' => 6 * 60 * 60 * 1000, // 6 hour chunks.
+					'interval' => 6 * HOUR_IN_SECONDS * 1000, // 6 hour chunks.
 					'extended_bounds' => [
-						'min' => milliseconds() - ( 7 * 24 * 60 * 60 * 1000 ),
-						'max' => milliseconds(),
+						'min' => Utils\milliseconds() - ( WEEK_IN_SECONDS * 1000 ),
+						'max' => Utils\milliseconds(),
```
joehoyle, line 363: Do these mean the below cache will never be hit?

inc/audiences/namespace.php
```diff
@@ -7,11 +7,7 @@
 
 namespace Altis\Analytics\Audiences;
 
-use function Altis\Analytics\Audiences\REST_API\get_audience_schema;
-use function Altis\Analytics\Utils\get_asset_url;
-use function Altis\Analytics\Utils\get_field_type;
-use function Altis\Analytics\Utils\milliseconds;
-use function Altis\Analytics\Utils\query;
```
joehoyle, line 14: this is a really badly named function `query` especially in a utils NS!

inc/audiences/namespace.php
rmccue: Good question. Probably, yes, but haven't changed from the original here, just refactoring. Best to leave that feedback on #36.

PR #44, opened 2020-04-03
Iterate on React for audience UI

rmccue:
Fixes a bunch of the React comments on #36.

### DIFF
--- a/src/audiences/edit/data/index.js
+++ b/src/audiences/edit/data/index.js
@@ -85,20 +85,20 @@ const selectors = {
 };
 
 const resolvers = {
-	* getFields() {
+	*getFields() {
 		const fields = yield actions.fetch( {
 			path: 'analytics/v1/audiences/fields',
 		} );
 		return actions.setFields( fields );
 	},
-	* getEstimate( audience ) {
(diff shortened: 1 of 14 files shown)


src/audiences/edit/data/index.js
```diff
@@ -57,128 +64,66 @@ const actions = {
...
+		return actions.setFields( fields );
+	},
+	*getEstimate( audience ) {
+		const audienceQuery = encodeURIComponent( JSON.stringify( audience ) );
+		const estimate = yield actions.fetch( {
+			path: `analytics/v1/audiences/estimate?audience=${ audienceQuery }`,
+		} );
+		return actions.addEstimate( audience, estimate );
+	},
+	*getPost( id ) {
```
joehoyle, line 101: Where is this used? I'm looking to understand if there's really value to using generators here as they are seldom used, especially by us.

rmccue: I believe this is used so we can essentially return multiple values from the function; not sure though

joehoyle: Mmm ok I guess we can leave that to the parent PR

PR #47, opened 2020-05-05
Support for providing arrays in attributes and metrics

### LINKED ISSUE #46: Allow arrays of string / floats for attributes & metrics
Event attributes can be arrays of strings and metrics can be arrays of floats, we should support this in the tracker script to allow for grouping values on more generic keys.

For instance to record categories for a pageview you might do something like this curreently:

```
Analytics.registerAttribute( `category_${categoryId}`, categorySlug );
```

It would be nicer to do:

```
Analytics.registerAttribute( 'category_ids', [ 1, 2, 3 ] );
```

This makes querying data with less starting information to build the query more practical.

roborourke:
Both AWS Pinpoint and Elasticsearch support arrays of the required datatype
   eg string or floats for attributes and metrics respectively. We were
   previsouly sanitising these to individual numbers and strings.

   The benefit of having array values for these means that rather than using
   attribute keys like `category_NN` where `NN` is the ID we could have an
   attribute containing all category IDs associated with the event eg.
   `category_ids: [1,2,3]`.

   Elasticsearch queries are unaffected by this change in the shape of the
   data as it does not differentiate between arrays and single values, only
   the data type is important.

   Fixes #46

### DIFF
--- a/src/analytics.js
+++ b/src/analytics.js
@@ -117,7 +117,7 @@ const prepareAttributes = async attributes => {
 };
 const prepareMetrics = async metrics => {
 	for ( const name in metrics ) {
-		metrics[ name ] = await prepareData( metrics[ name ], sanitiseMetric );
+		metrics[ name ] = await prepareData( metrics[ name ], sanitizeMetric );
 	}
 	return metrics;
 };
--- a/src/analytics.js
+++ b/src/analytics.js
@@ -105,7 +105,7 @@ const prepareData = async ( value, sanitiseCallback ) => {
 	if ( ! Array.isArray( value ) ) {
 		value = [ value ];
 	}
-	return value.map( val => sanitiseCallback( val ) );
+	return value.map( sanitizeCallback );
 };
 const sanitiseAttribute = value => value.toString();
 const sanitiseMetric = value => parseFloat( Number( value ) );
--- a/src/analytics.js
+++ b/src/analytics.js
@@ -98,35 +98,50 @@ const getSearchParams = () =>
 		} ),
 		{}
 	);
-const getAttributes = ( extra = {} ) =>
-	Object.entries( {
-		session: getSessionID(),
-		pageSession: pageSession,
-		url: window.location.origin + window.location.pathname,
-		host: window.location.hostname,
-		search: window.location.search,
-		hash: window.location.hash,
-		referer: document.referrer,
-		...getSearchParams(),
-		...utm,
-		...( Data.Attributes || {} ),
-		...extra,
-		...( _attributes || {} ),
-	} ).reduce( ( carry, [ name, value ] ) => ( {
-		...carry,
-		[ name ]: ( typeof value === 'function' ? value() : value ).toString(),
-	} ), {} );
-const getMetrics = ( extra = {} ) =>
-	Object.entries( {
-		elapsed: elapsed + ( Date.now() - start ),
-		scrollDepthMax,
-		scrollDepthNow,
-		...extra,
-		...( _metrics || {} ),
-	} ).reduce( ( carry, [ name, value ] ) => ( {
-		...carry,
-		[ name ]: Number( typeof value === 'function' ? value() : value ),
-	} ), {} );
+const prepareData = async ( value, sanitiseCallback ) => {
+	if ( typeof value === 'function' ) {
+		value = await value();
+	}
+	if ( ! ( value instanceof Array ) ) {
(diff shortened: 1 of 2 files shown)


roborourke:
Hm, thinking further this might need handling in the audience field data value retrieval too unless ES splits them up in aggregations.

src/analytics.js
```diff
@@ -98,35 +98,50 @@ const getSearchParams = () =>
...
-		[ name ]: Number( typeof value === 'function' ? value() : value ),
-	} ), {} );
+const prepareData = async ( value, sanitiseCallback ) => {
+	if ( typeof value === 'function' ) {
+		value = await value();
+	}
+	if ( ! Array.isArray( value ) ) {
+		value = [ value ];
+	}
+	return value.map( val => sanitiseCallback( val ) );
```
rmccue, line 108: Redundant arrow function, and also should use American English.

```suggestion
	return value.map( sanitizeCallback );
```

src/analytics.js
```diff
@@ -98,35 +98,50 @@ const getSearchParams = () =>
...
-		elapsed: elapsed + ( Date.now() - start ),
-		scrollDepthMax,
-		scrollDepthNow,
-		...extra,
-		...( _metrics || {} ),
-	} ).reduce( ( carry, [ name, value ] ) => ( {
-		...carry,
-		[ name ]: Number( typeof value === 'function' ? value() : value ),
-	} ), {} );
+const prepareData = async ( value, sanitiseCallback ) => {
```
rmccue, line 101:
```suggestion
const prepareData = async ( value, sanitizeCallback ) => {
```

src/analytics.js
```diff
@@ -98,35 +98,50 @@ const getSearchParams = () =>
...
+	if ( typeof value === 'function' ) {
+		value = await value();
+	}
+	if ( ! Array.isArray( value ) ) {
+		value = [ value ];
+	}
+	return value.map( val => sanitiseCallback( val ) );
+};
+const sanitiseAttribute = value => value.toString();
+const sanitiseMetric = value => parseFloat( Number( value ) );
```
rmccue, line 111:
```suggestion
const sanitizeAttribute = value => value.toString();
const sanitizeMetric = value => parseFloat( Number( value ) );
```

src/analytics.js
```diff
@@ -98,35 +98,50 @@ const getSearchParams = () =>
...
+const prepareAttributes = async attributes => {
+	for ( const name in attributes ) {
+		attributes[ name ] = await prepareData( attributes[ name ], sanitiseAttribute );
```
rmccue, line 114:
```suggestion
		attributes[ name ] = await prepareData( attributes[ name ], sanitizeAttribute );
```

src/analytics.js
```diff
@@ -98,35 +98,50 @@ const getSearchParams = () =>
...
+	}
+	return attributes;
```
rmccue, line 116: This will mutate `attributes`, that doesn't seem desired? Maybe instead:

```suggestion
	const sanitized = {};
	for ( const name in attributes ) {
		sanitized[ name ] = await prepareData( attributes[ name ], sanitizeAttribute );
	}
	return sanitized;
```

src/analytics.js
```diff
@@ -98,35 +98,50 @@ const getSearchParams = () =>
...
+};
+const prepareMetrics = async metrics => {
+	for ( const name in metrics ) {
+		metrics[ name ] = await prepareData( metrics[ name ], sanitiseMetric );
```
rmccue, line 120:
```suggestion
		metrics[ name ] = await prepareData( metrics[ name ], sanitizeMetric );
```

src/analytics.js
```diff
@@ -98,35 +98,50 @@ const getSearchParams = () =>
...
+	}
+	return metrics;
```
rmccue, line 122: As above, this will mutate.

```suggestion
	const prepared = {};
	for ( const name in metrics ) {
		prepared[ name ] = await prepareData( metrics[ name ], sanitizeMetric );
	}
	return prepared;
```

src/analytics.js
roborourke: I'd thought about it but felt it wasn't a big issue in this case but it could always lead to weirdnesses down the line. The upstream functions that call this one create new objects at least but better to have the behaviour contained 👍

src/analytics.js
roborourke: Deerp

PR #48, opened 2020-05-06
Make audience settings available client side

roborourke:
This exposes the audience configurations and parses them based on the current global attributes, metrics and endpoint data in order to assign the current page view to an audience.

The audiences are updated any time the endpoint data is updated or `registerAttribute()` or `registerMetric()` is called in order to keep them in sync with any custom global attributes or metrics that are derived async.

If the audiences change an event is triggered that allows external code to hook in and respond to the changes.

### DIFF
--- a/inc/audiences/namespace.php
+++ b/inc/audiences/namespace.php
@@ -98,6 +98,12 @@ function register_default_event_data_maps() {
 	register_field( 'attributes.utm_campaign', __( 'UTM Campaign', 'altis-analytics' ) );
 	register_field( 'attributes.utm_source', __( 'UTM Source', 'altis-analytics' ) );
 	register_field( 'attributes.utm_medium', __( 'UTM Medium', 'altis-analytics' ) );
+
+	// Time based parameters.
+	register_field( 'metrics.hour', __( 'Hour (0 - 23)', 'altis-analytics' ) );
+	register_field( 'metrics.day', __( 'Day (1 = Sunday, 7 = Saturday)', 'altis-analytics' ) );
+	register_field( 'metrics.month', __( 'Month (1 = January)', 'altis-analytics' ) );
+	register_field( 'metrics.year', __( 'Year', 'altis-analytics' ) );
 }
 
 /**
--- a/inc/audiences/namespace.php
+++ b/inc/audiences/namespace.php
@@ -9,6 +9,7 @@ namespace Altis\Analytics\Audiences;
 
 use Altis\Analytics\Utils;
 use WP_Post;
+use WP_Query;
 
 const COMPARISON_OPERATORS = [
 	'=',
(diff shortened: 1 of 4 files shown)


inc/audiences/namespace.php
```diff
@@ -755,3 +764,49 @@ function build_audience_query( array $audience ) : array {
 
 	return $groups_query;
 }
+
+/**
+ * Output audience configuration data as JSON.
```
joehoyle, line 769: Doesnt look like this `output`s? `Get ` seems better maybe?

inc/audiences/namespace.php
```diff
@@ -755,3 +764,49 @@ function build_audience_query( array $audience ) : array {
...
+	}
+
+	/**
+	 * Limits the number of active audiences there can be in use at any one time
+	 * on the front end.
+	 *
+	 * @param int $limit The number of audiences to fetch for use client side.
+	 */
+	$limit = apply_filters( 'altis.analytics.audiences.limit', 20 );
+	$limit = absint( $limit );
```
joehoyle, line 788: should this be `intval`, or do you specifically want to disallow `-1`?

inc/audiences/namespace.php
```diff
@@ -755,3 +764,49 @@ function build_audience_query( array $audience ) : array {
...
+
+	$audiences = new WP_Query( [
+		'post_type' => POST_TYPE,
+		'posts_per_page' => $limit,
+		'post_status' => 'publish',
```
joehoyle, line 793: Think this should have `no_found_rows => true`, or use `showposts`

src/analytics.js
```diff
@@ -244,6 +245,155 @@ const Analytics = {
 
 		return await Analytics.client;
 	},
+	audiences: [],
+	updateAudiences: async () => {
```
joehoyle, line 249: is this actually more like `fetchAudiences`? The naming now looks like "Add audiences to someone", but I thikn maybe it means "Update (fetch) the audiences"?

src/analytics.js
```diff
@@ -244,6 +245,155 @@ const Analytics = {
...
+
+		// Store derived field values for quicker lookups.
+		const fieldValues = {};
+
+		// Audience ID store.
+		const audienceIds = [];
+
+		// Map audience configurations to IDs.
+		for ( const aid in Audiences ) {
+			const { id, config } = Audiences[ aid ];
```
joehoyle, line 275: Could we pull this whole inner blocker into it's own function? Seems the job of "match the user endpoints (etc) to our Audience objects should be able to be self-contained in a function?

joehoyle: Ah looking below, perhaps this function does both: fetch the user's audiences and also trigger an "audiences updated" event if they have changed?

inc/audiences/namespace.php
roborourke: Oh whoops, it's not meant to that should have been updated

inc/audiences/namespace.php
roborourke: Yeah `-1` means 'all' in this context which we don't want

inc/audiences/namespace.php
roborourke: I was thinking about some sort of notification or indication that there's a limit on the number of active audiences, I'm not sure how best to approach that yet but probably in the UI so this change makes sense.

src/analytics.js
roborourke: Yeah I was thinking of maybe splitting it up but I'm not sure how much more readable it would be beyond less indentation. It was just easier for me to write it this way first.

src/analytics.js
roborourke: Not quite, it looks at the available data (attributes, metrics and endpoint) then figures what audiences the current page session belongs to. So it's not really "fetching" anything, at least in the HTTP sense.

Ultimately it sets the `audiences` array on the line above which can then be retrieved so I think "update" makes the most sense.

PR #52, opened 2020-05-07
Add additional default audience segment options

### LINKED ISSUE #49: Add ability to segment by recurring visits
It'll be useful for experience blocks and personalisation implementations to be able to create segments based on whether someone is a recurring visitor or not.

One possible approach would be to record an endpoint metric for the following:

1. number of sessions
2. number of pages visited

This could then easily be used when defining audiences.

### LINKED ISSUE #50: Add ability to segment by time based values
This should include:

- hour of day
- day of week
- month
- year

It's unlikely that year will be needed but we can include for good measure.

These should be added as default global `metrics` in `analytics.js` so that numerical range based queries can be carried out on them.

### LINKED ISSUE #51: Add default segmentation by UTM query parameters
We are already tracking UTM parameters specifically with every recorded event however we should also expose these by default in the audience builder UI.

- `utm_campaign`
- `utm_source`
- `utm_medium`

Can also add the relatively uncommon:

- `utm_term`
- `utm_content`

roborourke:
- Time based audiences
- Page view & session count based audiences
- UTM parameter audiences

Fixes #49
Fixes #50
Fixes #51

### DIFF
--- a/src/analytics.js
+++ b/src/analytics.js
@@ -473,6 +473,25 @@ const Analytics = {
 		endpoint.Attributes = await prepareAttributes( endpoint.Attributes );
 		endpoint.Metrics = await prepareMetrics( endpoint.Metrics );
 
+		// Add session and page view counts to endpoint.
+		if ( ! endpoint.Attributes.lastSession ) {
+			endpoint.Attributes.lastSession = [ getSessionID() ];
+			endpoint.Attributes.lastPageSession = [ pageSession ];
+			endpoint.Metrics.sessions = [1.0];
+			endpoint.Metrics.pageViews = [1.0];
+		} else {
+			// Increment sessions.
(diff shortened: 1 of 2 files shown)


src/analytics.js
```diff
@@ -142,6 +144,10 @@ const getMetrics = ( extra = {} ) => ( {
 	elapsed: elapsed + ( Date.now() - start ),
 	scrollDepthMax,
 	scrollDepthNow,
+	hour: new Date().getHours(),
+	day: new Date().getDay() + 1,
```
rmccue, line 148: Month I get, but why offset this by 1?

roborourke: Good q, I figured if I did it for month I should do it for day too, otherwise it's a little inconsistent and we're exposing the idea of 0 indexed lists to users. So Sunday is 1, etc, etc

PR #56, opened 2020-05-15
Replace default audience post type UI with React app

roborourke:
This completely swaps out the default editing UI for a dynamic table with sorting capabilities.

In addition an audience picker component that pops up the audience editing interface in a modal is exported that can be enqueued and used by other plugins / scripts.

PR #60, opened 2020-05-28
Add ability to preview audiences

rmccue:
Branched from #56.

Adds an audience selector drop-down to the admin bar for users who can `edit_posts`:

Screen Shot 2020-05-27 at 18 33 56

Few things to consider:
* Do we want a custom cap for this instead?
* Should it avoid React? We run the risk of conflicting with React app frontends.

### DIFF
--- a/inc/preview/namespace.php
+++ b/inc/preview/namespace.php
@@ -1,0 +1,103 @@
+<?php
+/**
+ * Audience preview selector.
+ *
+ * @package altis-analytics
+ */
+
+namespace Altis\Analytics\Preview;
+
+use Altis\Analytics;
+use Altis\Analytics\Audiences;
+use Altis\Analytics\Utils;
+
+/**
+ * Bootstrap Title AB Tests Feature.
+ *
+ * @return void
+ */
+function setup() : void {
+	add_action( 'wp_enqueue_scripts', __NAMESPACE__ . '\\enqueue_assets' );
+	add_action( 'admin_bar_menu', __NAMESPACE__ . '\\add_menu_item', 100 );
+}
+
+/**
+ * Should we show the Audience preview selector?
+ *
+ * @return bool True to show the preview selector, false otherwise.
+ */
+function should_show_selector() : bool {
+	return ( ! is_admin() ) && current_user_can( 'edit_posts' );
+}
+
+/**
+ * Enqueue the preview selector assets.
+ *
+ * @return void
+ */
+function enqueue_assets() : void {
+	if ( ! should_show_selector() ) {
+		return;
+	}
+
+	wp_enqueue_style(
+		'altis-analytics-preview',
+		plugins_url( 'src/preview/index.css', Analytics\ROOT_DIR . '/plugin.php' ),
+		[],
+		'2020-05-27-1'
+	);
+
+	// Enqueue the script and pass through data.
+	wp_enqueue_script(
+		'altis-analytics-preview',
+		Utils\get_asset_url( 'preview.js' ),
+		[
+			'altis-analytics',
+			'wp-element',
+		],
+		null,
+		true
+	);
+	wp_localize_script( 'altis-analytics-preview', 'AltisExperimentsPreview', get_script_data() );
+}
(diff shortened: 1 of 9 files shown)


rmccue:
Marking as ready-for-review; Rob noted we're actually using React in notifications already, so that's not an issue for now.

Custom cap is still worth considering I think.

src/preview/components/Selector.js
```diff
@@ -0,0 +1,55 @@
...
+			<a
+				aria-haspopup="true"
+				className="ab-item"
+				href="#qm-overview"
+			>
+				Audiences
+			</a>
+			<div className="ab-sub-wrapper">
+				<ul
+					className="ab-submenu"
```
roborourke, line 34: Guess this class name can go on one line too

src/preview/components/Selector.js
```diff
@@ -0,0 +1,55 @@
...
+				>
+					{ audiences.map( audience => (
+						<li
+							key={ audience.id }
+							className="aa-preview-item"
+						>
+							<a
+								className={ `ab-item ${ isSelected( audience.id ) ? 'altis-analytics-preview-selected' : '' }` }
+								href="#"
+								role="button"
```
roborourke, line 44: Would an actual button would require some extra styling to match the admin bar links if it has the same class name?

src/analytics.js
```diff
@@ -653,6 +666,7 @@ window.addEventListener( 'beforeunload', () => {
 window.Altis.Analytics.updateEndpoint = Analytics.updateEndpoint;
 window.Altis.Analytics.getEndpoint = Analytics.getEndpoint;
 window.Altis.Analytics.getAudiences = Analytics.getAudiences;
+window.Altis.Analytics.overrideAudiences = Analytics.overrideAudiences;
```
roborourke, line 669: Why would we always want this to be available in user land? I was thinking it's better to replace `window.Altis.Analytics.getAudiences`. The main reason is that it's an admin/editorial use-case rather than something needed for a regular visitor. I know it's not a lot of code but I'd like to keep the main `analytics.js` file size as small as possible.

Still, if it's a pain to get around then no worries. We can address the overall state of this file in a future iteration.

inc/preview/namespace.php
```diff
@@ -0,0 +1,103 @@
...
+	add_action( 'admin_bar_menu', __NAMESPACE__ . '\\add_menu_item', 100 );
+}
+
+/**
+ * Should we show the Audience preview selector?
+ *
+ * @return bool True to show the preview selector, false otherwise.
+ */
+function should_show_selector() : bool {
+	return ( ! is_admin() ) && current_user_can( 'edit_posts' );
```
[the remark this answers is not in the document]
roborourke: Any reason for the brackets? Removed an or statement?
```suggestion
	return ! is_admin() && current_user_can( 'edit_posts' );
```

src/analytics.js
roborourke: For reference you could do something like:

```js
var getRealAudiences = Altis.Analytics.getAudiences;
Altis.Analytics.getAudiences = function () {
     // Return overrides if set
     if ( overrides ) {
         return overrides;
     }
     return getRealAudiences();
};
```

rmccue: Good Q; I think it is useful, and it was basically "free", but can change if you'd prefer not to expose as a public API.

src/preview/components/Selector.js
rmccue: Yeah, it'd require a *lot* of extra styling; the admin bar isn't really built for this level of interactivity.

inc/preview/namespace.php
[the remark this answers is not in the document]
rmccue: Oh yeah I added that because I always forget whether ! or && is evaluated first, but can remove

src/analytics.js
roborourke: I'd prefer not to but lets leave it for now and just not document it. Want to get this over the line really.

PR #62, opened 2020-06-01
Replace default audience post type UI with React app

roborourke:
This completely swaps out the default editing UI for a dynamic table with sorting capabilities.

In addition an audience picker component that pops up the audience editing interface in a modal is exported that can be enqueued and used by other plugins / scripts.


roborourke:
Not sure what I did here, hub cli fail I think.

PR #64, opened 2020-06-05
Tagging 2.0.0

roborourke:
Trigger 2.0.0 build and release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "altis-analytics",
-  "version": "2.0.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "altis-analytics",
-  "version": "0.0.1",
+  "version": "2.0.0",
```
rmccue, line 3: This should probably be `private: true` anyway?

roborourke: TIL

PR #83, opened 2020-06-16
Fix audience editor button styling

roborourke:


src/audiences/ui/components/rule.js
```diff
@@ -23,7 +23,7 @@ const StyledRule = styled.div`
 
 	.audience-editor__rule-operator,
 	button {
-		flex 0;
+		flex: none;
```
rmccue, line 26: Actually, I think this is invalid. `flex: 0` is shorthand for `flex-grow: 0`, but I don't think `flex: none` is a thing.

roborourke: Totes valid m8 https://developer.mozilla.org/en-US/docs/Web/CSS/flex

rmccue: Aha, equivalent to `flex: 0 0 auto`. Neat!

PR #89, opened 2020-06-23
Dont show audience preview if no audiences

### LINKED ISSUE #53: Ensure estimate rules with an empty field name are ignored
There can be some strange behaviour while putting together audience estimates resulting in percentages greater than 100% showing or even 100% showing before any field & value has been selected.

Any rules with any empty field parameter should be ignored.

roborourke:
Fixes #53

### DIFF
--- a/inc/preview/namespace.php
+++ b/inc/preview/namespace.php
@@ -27,7 +27,7 @@ function setup() : void {
  * @return bool True to show the preview selector, false otherwise.
  */
 function should_show_selector() : bool {
-	return is_admin_bar_showing() && ! is_admin() && current_user_can( 'edit_posts' ) && count( Audiences\query_audiences() ) > 0;
+	return is_admin_bar_showing() && ! is_admin() && current_user_can( 'edit_posts' );
 }
 
 /**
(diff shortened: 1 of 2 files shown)


inc/preview/namespace.php
```diff
@@ -27,7 +27,7 @@ function setup() : void {
  * @return bool True to show the preview selector, false otherwise.
  */
 function should_show_selector() : bool {
-	return is_admin_bar_showing() && ! is_admin() && current_user_can( 'edit_posts' );
+	return is_admin_bar_showing() && ! is_admin() && current_user_can( 'edit_posts' ) && count( Audiences\query_audiences() ) > 0;
```
rmccue, line 30: This solves the problem, but I wonder if it should have explanatory text instead so we're encouraging more use of audiences?

roborourke: Hmm. Could have a link to the audience editor admin page as well perhaps. I don't know really, do you want to update this with that suggestion? If you're not sure lets just merge this for now and create a new issue for your suggestion.

rmccue:
We decided to show a message here instead; I can tackle.

PR #107, opened 2020-07-27
Add fallback for caps

rmccue:
Adds a fallback for the capabilities to the page capability. The caps can still be explicitly set or removed on roles as desired, and a constant is added to disable the fallback functionality.

I chose this over simply adding the caps to the roles as this way it grants the caps to custom roles as well, and doesn't require modifying any role objects as part of a DB migration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,12 @@ Allows you to define the Elasticsearch server URL directly.
 
 Define as true to enable logging queries to the error log.
 
+**`ALTIS_ANALYTICS_FALLBACK_CAPS`**
+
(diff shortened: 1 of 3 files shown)


roborourke:
Why choose page caps for this? I think post caps might make more sense, allows content creators to make audiences but not edit others audiences etc according to their role. I kinda get that personalised content makes more sense on static pages but we don't really know that for certain.

roborourke:
Oh actually can you bump the version number in `plugin.php` too? Can backport and release a v2 branch update then.

rmccue:
> Why choose page caps for this? I think post caps might make more sense, allows content creators to make audiences but not edit others audiences etc according to their role. I kinda get that personalised content makes more sense on static pages but we don't really know that for certain.

In this case I was thinking more about the structure of a site, and treating pages as the main form of content rather than posts. Same reasoning as you noted from there.

PR #110, opened 2020-08-05
Don't schedule events during install

roborourke:
Fixes an error on WP install where the options table hasn't been created on the `init` action yet.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -40,7 +40,7 @@ function schedule_events() {
 	if ( defined( 'WP_INSTALLING' ) && WP_INSTALLING ) {
 		return;
 	}
-	
+
 	if ( function_exists( 'is_main_site' ) && ! is_main_site() ) {
 		return;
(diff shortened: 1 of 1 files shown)


inc/namespace.php
```diff
@@ -37,6 +37,10 @@ function setup() {
  * Schedule common maintenance tasks.
  */
 function schedule_events() {
+	if ( defined( 'WP_INSTALLING' ) && WP_INSTALLING ) {
```
rmccue, line 40: Just to double check, should this be `WP_INITIAL_INSTALL` instead?

roborourke: That only exists with Altis core so this is fine here. But now I think about it this might be getting added for every site. May need a follow up. Not sure if there's a solid solution for network level cron.

rmccue: Solution is to check if you're on the main site and only schedule there.

PR #118, opened 2020-09-08
Release 2.4.0

roborourke:
WP 5.5 support version


joehoyle (approved):
Just checking if we want to backport-- is it backwards compatible?

roborourke:
@joehoyle we're tracking the 2.3.x version currently so this update won't affect existing installs

PR #122, opened 2020-11-10
Add S3 cleanup routine

### LINKED ISSUE #111: Add S3 data lake clean up to data deletion routine
Currently we have a default clean routine that clears elasticsearch indexes older than 14 days. AWS Pinpoint itself will delete data after 90 days so we should also clean up the data backed up to S3 using 90 days as a default.

roborourke:
Defaults to 90 days to match AWS Pinpoint after which time backup data is removed from S3 storage.

Fixes #111

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ Define as false to disable the capability fallback to page capabilities.
 
 **`ALTIS_ANALYTICS_PINPOINT_BUCKET_ARN`**
 
-If you have the Pinpoint Kinesis Firehose set to back up data to S3 then defining this constant allows for data to be cleaned from the backups periodically too.
+If you have the Pinpoint Kinesis Firehose set to back up data to S3, defining this constant allows for data to be cleaned from the backups periodically.
 
 **`ALTIS_ANALYTICS_PINPOINT_BUCKET_REGION`**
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -108,6 +108,14 @@ By default, Altis Analytics will grant any role who can edit pages the ability t
 
 Define as false to disable the capability fallback to page capabilities.
 
+**`ALTIS_ANALYTICS_PINPOINT_BUCKET_ARN`**
+
+If you have the Pinpoint Kinesis Firehose set to back up data to S3 then defining this constant allows for data to be cleaned from the backups periodically too.
```
jazzsequence, line 113: This should have some commas. 

```suggestion
If you have the Pinpoint Kinesis Firehose set to back up data to S3, defining this constant allows for data to be cleaned from the backups periodically, too.
```

README.md
```diff
@@ -142,6 +150,10 @@ Filter the maximum number of days to keep real time stats available for. The def
 
 Insights and aggregated analytics data can be calculated, updated and stored in the database in cases where you wish to retain information for longer periods of time such as number of page views.
 
+**`altis.analytics.max_s3_backup_age <int>`**
+
+Filter the maximum number of days to keep backup data for. The default number of days is 90 in accordance with AWS Pinpoint, after which time data is removed. This is important for streamlining your user's privacy. There is no upper limit on this value however you should make sure any long term data storage is explained to users when opting in to tracking.
```
jazzsequence, line 155: The comma after AWS Pinpoint feels weird. I think it should either be removed, or another comma should be added after `90`.

`user's` should be `users'` to show plural user possessive.

& there should be a comma after `however`.

```suggestion
Filter the maximum number of days to keep backup data for. The default number of days is 90, in accordance with AWS Pinpoint, after which time data is removed. This is important for streamlining your users' privacy. There is no upper limit on this value, however you should make sure any long term data storage is explained to users when opting in to tracking.
```

jazzsequence (approved):
No real questions here on the code, not being intimately familiar with the underlying AWS stuff yet, but it's all documented well and makes sense logically. A couple comments on the documentation additions that should be fixed before merge, but otherwise 👍.

README.md
roborourke: Hm, the comma just before the last word in the sentence is weird to me. Saying the sentence out loud doesn't create a natural pause there. I think we can just remove the word "too".

PR #128, opened 2020-12-09
Add data export endpoint

roborourke:
Add the endpoint `/wp-json/analytics/v1/events/YYYY-MM-DD` - this allows exporting data for the given day.

Pagination info is delivered via response headers.

Still need to add docs around using app passwords for auth.

### DIFF
--- a/inc/export/class-endpoint.php
+++ b/inc/export/class-endpoint.php
@@ -1,11 +1,7 @@
 <?php
 /**
  * REST API endpoint to retrieve all the Altis Analytics data for a particular date.
- *
- * Displays 5000 items at once, after that there's pagination.
  */
-
-declare( strict_types=1 );
 
 namespace Altis\Analytics\Export;
 
@@ -47,12 +43,12 @@ class Endpoint {
 						'description' => 'How many records to return per chunk.',
 						'type' => 'number',
 						'default' => 2000,
-                    ],
-                    'format' => [
-                        'description' => 'The data format to get results in, one of json or csv.',
-                        'type' => 'string',
-                        'enum' => [ 'json', 'csv' ],
-                        'default' => 'json',
+					],
+					'format' => [
+						'description' => 'The data format to get results in, one of json or csv.',
+						'type' => 'string',
+						'enum' => [ 'json', 'csv' ],
+						'default' => 'json',
 					],
 				],
 				[
@@ -205,6 +201,7 @@ class Endpoint {
 			$events = wp_list_pluck( $results['hits']['hits'], '_source' );
 
 			if ( $format === 'json' ) {
+				// phpcs:ignore HM.Security.EscapeOutput.OutputNotEscaped
 				echo trim( wp_json_encode( $events, JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE ), '[]' ) . "\n";
 				// Add a comma in between result sets except for the last page.
 				echo $page === $total_pages - 1 ? '' : ',';
@@ -225,7 +222,7 @@ class Endpoint {
 
 		// Close out the JSON array.
 		if ( $format === 'json' ) {
-			echo "]";
+			echo ']';
 		}
 
 		// Cleanup resource heavy scroll query.
@@ -237,6 +234,7 @@ class Endpoint {
 	/**
 	 * Removes any existing Elasticsearch search scrolls.
 	 *
+	 * @param array $scroll_ids A list of scroll IDs to be removed.
 	 * @return void
 	 */
 	private function clear_search_scroll( array $scroll_ids ) {
--- a/inc/export/class-endpoint.php
+++ b/inc/export/class-endpoint.php
@@ -1,0 +1,323 @@
+<?php
+/**
+ * REST API endpoint to retrieve all the Altis Analytics data for a particular date.
+ *
+ * Displays 5000 items at once, after that there's pagination.
+ */
+
+declare( strict_types=1 );
+
+namespace Altis\Analytics\Export;
+
+use Altis\Analytics\Utils;
+use DateTime;
+use Exception;
+use WP_Error;
+use WP_REST_Request;
+use WP_REST_Server;
+
+use function HM\GTM\flatten_array as GTMFlatten_array;
+
+/**
+ * Class Data_Endpoint
+ *
+ * @package RN\Altis_Analytics_Customisations
+ */
+class Endpoint {
+	/**
+	 * Add hooks.
+	 */
+	public function bootstrap() : void {
+		add_action( 'rest_api_init', [ $this, 'register_routes' ] );
+		add_filter( 'rest_pre_serve_request', [ $this, 'pre_serve_request' ], 10, 4 );
+	}
+
+	/**
+	 * Registers the route.
+	 */
+	public function register_routes() : void {
+		register_rest_route(
+			'analytics/v1',
+			'export/(?P<date>\d{4}-\d{2}-\d{2})',
+			[
+				'args'   => [
+					'date' => [
+						'description' => 'Day to retrieve the analytics data from.',
+						'type' => 'date',
+						'sanitize_callback' => 'sanitize_text_field',
+					],
+					'page' => [
+						'description'       => 'Current page',
+						'type'              => 'integer',
+						'default'           => 1,
+						'sanitize_callback' => 'absint',
+						'validate_callback' => 'rest_validate_request_arg',
+						'minimum'           => 1,
+					],
+					'per_page' => [
+						'description' => 'How many records to return per page.',
+						'type' => 'number',
+						'sanitize_callback' => 'absint',
+                    ],
+                    'format' => [
+                        'description' => 'The data format to get results in, one of json or csv.',
+                        'type' => 'string',
+                        'enum' => [ 'json', 'csv' ],
+                        'default' => 'json',
+                    ],
+				],
+				[
+					'methods' => WP_REST_Server::READABLE,
+					'callback' => [ $this, 'get_items' ],
+					'permission_callback' => [ $this, 'get_item_permissions_check' ],
+				],
+			]
+		);
+	}
+
+	/**
+	 * Return the items for this endpoint.
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 *
+	 * @return WP_REST_Response|WP_Error Response object.
+	 */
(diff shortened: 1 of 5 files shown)


inc/export/class-endpoint.php
```diff
@@ -0,0 +1,323 @@
...
+						'description' => 'Day to retrieve the analytics data from.',
+						'type' => 'date',
+						'sanitize_callback' => 'sanitize_text_field',
+					],
+					'page' => [
+						'description'       => 'Current page',
+						'type'              => 'integer',
+						'default'           => 1,
+						'sanitize_callback' => 'absint',
+						'validate_callback' => 'rest_validate_request_arg',
```
joehoyle, line 54: these are not needed when passing correct `type` -- I've seen this pattern some a few people though, do you know where it comes from?

inc/export/class-endpoint.php
```diff
@@ -0,0 +1,323 @@
...
+	 * @return true
+	 */
+	public function pre_serve_request( bool $served, $result, WP_REST_Request $request, WP_REST_Server $server ) : bool {
+		$params = $request->get_params();
+
+		if ( ! preg_match( '#/analytics/v1/export/(?P<date>\d{4}-\d{2}-\d{2})#', $request->get_route() ) || 'GET' !== $request->get_method() ) {
+			return $served;
+		}
+
+		if ( ! isset( $params['format'] ) || 'csv' !== $params['format'] ) {
```
joehoyle, line 183: does it make sense to read from `Accept` header too?

inc/export/class-endpoint.php
```diff
@@ -0,0 +1,323 @@
...
+	}
+
+	/**
+	 * Lock down the endpoint to only allow administrator users authenticated via application password to access the
+	 * date.
+	 *
+	 * @return bool True if the request has read access for the item, otherwise false.
+	 */
+	public function get_item_permissions_check() {
+		return current_user_can( 'manage_options' );
```
joehoyle, line 321: Assuming we don't have any meta caps in analytics?

inc/export/class-endpoint.php
```diff
@@ -0,0 +1,323 @@
...
+		add_filter( 'rest_pre_serve_request', [ $this, 'pre_serve_request' ], 10, 4 );
+	}
+
+	/**
+	 * Registers the route.
+	 */
+	public function register_routes() : void {
+		register_rest_route(
+			'analytics/v1',
+			'export/(?P<date>\d{4}-\d{2}-\d{2})',
```
joehoyle, line 41: isn't an API always "exporting"? what's the data type? `events`?

joehoyle (commented):
Looking good.

I don't think it's in here, but it might be valuable to map some ID fields to slug. E>g. audience IDs or Post IDs are pretty useless when exporting data. Having the audience name in the CSV along with it's ID would be valuable.

inc/export/class-endpoint.php
roborourke: This was from the original code, not sure on the pattern itself. Maybe @fklein-lu would have more information

inc/export/class-endpoint.php
roborourke: Not for analytics directly but we could add an alternative capability here or at least a filter in the meantime.

inc/export/class-endpoint.php
roborourke: GET requests are I suppose. Yeah `events` is a fair description of the type 👍

inc/export/class-endpoint.php
roborourke: I'm not sure, I thought `Accept` headers could contain multiple mime types. Can do though, shouldn't be hard to add.

rmccue: `Accept` can contain multiple MIME types, and it's a bit of a pain to parse, but it's likely worth doing. [RFC 7231 §5.3.2](https://tools.ietf.org/html/rfc7231#section-5.3.2) is the relevant part.

To summarise from my memory: `Accept` is a comma-separated list of multiple values (as all† HTTP headers are). Each value is specified as `<type>/<subtype>` which can optionally have a wildcard (`*`) as the subtype or type and subtype, and can be followed with a semi-colon and then an arbitrary number of `key=value` parameter pairs separated by semicolons. The main one of those parameters is `q` which indicates the weight (i.e. priority) with 1 being the highest.

i.e. `Accept: text/html, application/xhtml+xml, application/xml;q=0.9, image/webp, */*;q=0.8` is
* Give me `text/html`, `application/xhtml+xml`, or `image/webp` with highest weight (q=1)
* Give me `application/xml` as the fallback, with weight 0.9
* Failing those, give me anything with `q=0.8`

roborourke:
@joehoyle re.

> I don't think it's in here, but it might be valuable to map some ID fields to slug. E>g. audience IDs or Post IDs are pretty useless when exporting data. Having the audience name in the CSV along with it's ID would be valuable.

I'm not sure if this is the best place to do that. I consciously opted not to use data points that are quite as mutable as post slugs by default but to be fair I was mainly thinking about usage within WP. I would be tempted to recommend as far as Data Studio usage is concerned creating additional data sources, eg. get a CSV of audiences, posts etc if you want to enrich the data. You do at least get the URL in the attributes.

Otherwise I'd aim to handle extending the default data added separately to this PR. In the interim you can filter the default data passed from the server side as well if you want more info in there.

inc/export/class-endpoint.php
roborourke: Changed `export` to `events`

inc/export/class-endpoint.php
roborourke: Added a filter for now but we can come back to this. I feel like we'd need a solid solution for role & cap management first.

inc/export/class-endpoint.php
roborourke: OK, added. Seems to work ok. Cheers @rmccue !

joehoyle:
Ok - in terms of what I'm looking for, for the integrate into data studio etc, I'd expect the following:

- Audience names so I don't have to map an internal ID to more complex combinations of CSVs etc
- Significantly higher event limits, I don't want to have to write a program to loop over pagination. (if you want prior work: I wrote a css/json streaming output for vast amounts of data on vantage-backend for the access logs endpoint). 

It might be that this PR doesn't need to do those things, 1 sounds like something that could come later. 2 does seem like an architectural decision though, that might be better to approach now. If you are streaming data with appropriate nginx buffer headers etc, you should be able to keep the connection open for more than 60 seconds and as long as you don't have memory leaks you shouldn't run out of memory.

Right now, this seems mostly like a REST API for events.. which is pretty expected and good n all... but I think there's a gap between that and the feature / use-case of being able to export data for use in other tools. The "include all the data you need, not just internal IDs" comes to that question.

From the code side, I think what you have makes a good rest api endpoint. I'll defer to @rmccue on a final review. My feedback is mostly around the use-case and usefulness.

roborourke:
I was thinking streaming data would be a better approach, we’re a bit stuck in terms of how Elasticsearch and paging works though. I’ll see how difficult it is.

We can probably add the audience names to the endpoint data by default but that would be done in the JS. I guess my thought was that those audiences might be defined in the BI tool too.

inc/export/class-endpoint.php
```diff
@@ -0,0 +1,292 @@
...
+
+			foreach ( $index_mappings as $mapping ) {
+				$fields = array_merge( $fields, array_keys( $mapping['mappings'] ) );
+				$fields = array_unique( $fields );
+			}
+
+			sort( $fields );
+
+			// Write columns headings line.
+			fputcsv( fopen( 'php://output', 'w' ), $fields ) . "\n";
```
rmccue, line 169: You're reopening output handles each time here; you should open the handle once. Also, after each call here, you should also call `fflush()`

roborourke: Okey doke, might need some updates where this pattern is used elsewhere 😬

roborourke:
Ok, streaming appears to work locally at least. I haven't tested beyond 20k documents but it seems to be able to export them in less than 30 seconds anyway. Requesting the CSV format seems to automatically trigger a download in the browser anyway so that's good. We could just add a link to the appropriate URL in the backend somewhere in future.

rmccue:
Can we test this early on production infrastructure? I want to ensure we're checking compatibility early.

roborourke:
@rmccue we can merge this, make a new release of this plugin and add it to a new preview release of the Altis Analytics module

roborourke:
Deployed to platform-test but the kinesis firehose doesn't seem to be delivering to the new ES domain yet.

roborourke:
Had an issue with cloud not allowing accept headers. Is there a way to allow that for a particular endpoint or is that a CloudFront limitation we have between the load balancer and the web server?

joehoyle (commented):
@roborourke wondering if / why you don't need to set the accel-buffering header to make Nginx not buffer the response (see https://github.com/humanmade/vantage-backend/blob/6a16494d60a7a6e0d622a376f3069604ae8788ff/content/mu-plugins/hm-stack-core/inc/namespace.php#L252)

Do we not need that, or were you testing the streaming some other way?

inc/utils/namespace.php
```diff
@@ -109,16 +109,45 @@ function get_elasticsearch_url() : string {
...
 }
 
+/**
+ * Retrieve the elasticsearch version.
+ *
+ * Return the version string if found otherwise 'unknown'.
+ *
+ * @return string
+ */
+function get_elasticsearch_version() : string {
```
joehoyle, line 119: Could we have this set by cloud maybe in a constant too? It's a bit of a pain when we're pinging off requests to elastic search on potentially unrelated page loads.

roborourke:
Ok! This works on cloud now with an application password, we can maybe look at adding a link somewhere in the admin to fetch it but not sure how best to authenticate from a link right now.

roborourke:
@joehoyle the `X-Accel-Buffering` header is set in the `get_items()` method. I just took it from the example you sent me. Is there more I need to add to make it work?

inc/utils/namespace.php
roborourke: Yeah if we update TAS to set it as an environment variable we could use that if it exists. Can you create an issue on the cloud module for that?

roborourke: Added an issue to the terraform repo instead as that's where we'd get the info from.

joehoyle (approved):
Fandabidozi

PR #131, opened 2020-12-18
Integrate WP Consent API

roborourke:
This adds support for the Altis Consent API and automatically integrates it with the analytics layer. We need to do some more QA regarding the behaviour of XBs and AB tests and how they degrade gracefully. Right now XBs display nothing and AB tests just show the default fallback content.

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -224,9 +224,19 @@ function enqueue_scripts() {
 	/**
 	 * Filters the consent cookie prefix to integrate with the WordPress Consent API.
 	 *
+	 * If the value is `null` then no special consent handling is done and Analytics will
+	 * run as normal.
+	 *
 	 * @param string|null $cookie_prefix The consent cookie prefix or false to skip consent check.
 	 */
-	$consent_cookie = apply_filters( 'altis.analytics.consent_cookie_prefix', null );
+	$consent_cookie_prefix = apply_filters( 'altis.analytics.consent_cookie_prefix', null );
+
+	/**
+	 * Filters the consent cookie prefix to integrate with the WordPress Consent API.
+	 *
+	 * @param string $consent_category The consent category to allow analytics to run with.
+	 */
+	$consent_category = apply_filters( 'altis.analytics.consent_category', 'statistics' );
 
 	wp_enqueue_script( 'altis-analytics', Utils\get_asset_url( 'analytics.js' ), [], null, false );
 	wp_add_inline_script(
@@ -243,7 +253,10 @@ function enqueue_scripts() {
 			wp_json_encode(
 				[
 					'Ready' => false,
-					'ConsentCookie' => $consent_cookie,
+					'Consent' => [
+						'CookiePrefix' => $consent_cookie_prefix,
+						'Category' => $consent_category,
+					],
 					'Config' => [
 						'PinpointId' => defined( 'ALTIS_ANALYTICS_PINPOINT_ID' ) ? ALTIS_ANALYTICS_PINPOINT_ID : null,
 						'PinpointRegion' => defined( 'ALTIS_ANALYTICS_PINPOINT_REGION' ) ? ALTIS_ANALYTICS_PINPOINT_REGION : null,
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -222,6 +222,13 @@ function enqueue_scripts() {
 	 */
 	$noop = (bool) apply_filters( 'altis.analytics.noop', false );
 
+	/**
+	 * Filters the consent cookie prefix to integrate with the WordPress Consent API.
+	 *
+	 * @param string|false $cookie_prefix The consent cookie prefix or false to skip consent check.
+	 */
+	$consent_cookie = apply_filters( 'altis.analytics.consent_cookie_prefix', false );
+
 	wp_enqueue_script( 'altis-analytics', Utils\get_asset_url( 'analytics.js' ), [], null, false );
 	wp_add_inline_script(
 		'altis-analytics',
@@ -237,7 +244,7 @@ function enqueue_scripts() {
 			wp_json_encode(
 				[
 					'Ready' => false,
-					'UseConsent' => function_exists( 'Altis\\Consent\\should_display_banner' ) && Consent\should_display_banner(),
+					'ConsentCookie' => $consent_cookie,
 					'Config' => [
 						'PinpointId' => defined( 'ALTIS_ANALYTICS_PINPOINT_ID' ) ? ALTIS_ANALYTICS_PINPOINT_ID : null,
 						'PinpointRegion' => defined( 'ALTIS_ANALYTICS_PINPOINT_REGION' ) ? ALTIS_ANALYTICS_PINPOINT_REGION : null,
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -222,21 +222,18 @@ function enqueue_scripts() {
 	$noop = (bool) apply_filters( 'altis.analytics.noop', false );
 
 	/**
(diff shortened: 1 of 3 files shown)


src/analytics.js
```diff
@@ -804,11 +762,86 @@ window.Altis.Analytics.registerAttribute = ( name, value ) => {
...
+}
+
+// Check if this is anonymous or not by looking for endpoint user ID.
+const isAnonymous = ! ( Data.Endpoint && Data.Endpoint.User && Data.Endpoint.User.UserId );
+const consentType = 'statistics' + ( isAnonymous ? '-anonymous' : '' );
+
+// Check Altis Consent feature is in use.
+if ( Altis.Analytics.UseConsent ) {
+	// Check cookie directly for an early match.
+	if ( document.cookie.match( `_altis_consent_${ consentType }=allow` ) ) {
```
ivankristianto, line 834: Since prefix `_altis_consent_` is filterable [here](https://github.com/humanmade/altis-consent/blob/main/inc/namespace.php#L36), Can we use [Altis.Consent.getCookie](https://github.com/humanmade/altis-consent/blob/main/assets/js/main.js#L67) instead?

roborourke: This script is loaded really early so we can’t rely on that, but good call, we can probably use the same filter and pass it through.

inc/namespace.php
```diff
@@ -221,20 +221,29 @@ function enqueue_scripts() {
 	 */
 	$noop = (bool) apply_filters( 'altis.analytics.noop', false );
 
+	/**
+	 * Filters the consent cookie prefix to integrate with the WordPress Consent API.
+	 *
+	 * @param string|null $cookie_prefix The consent cookie prefix or false to skip consent check.
+	 */
+	$consent_cookie = apply_filters( 'altis.analytics.consent_cookie_prefix', null );
```
jazzsequence, line 229: I would use the default `_altis_consent` rather than `null` here. I wouldn't want the default (null) to be passed through and then not detect the cookies (but maybe I'm mis-reading this?).

src/analytics.js
```diff
@@ -804,11 +762,86 @@ window.Altis.Analytics.registerAttribute = ( name, value ) => {
...
+
+	// Fire a ready event once userland API has been exported.
+	Altis.Analytics.Ready = true;
+	const readyEvent = new CustomEvent( 'altis.analytics.ready' );
+	window.dispatchEvent( readyEvent );
+}
+
+// Check if this is anonymous or not by looking for endpoint user ID.
+const isAnonymous = ! ( Data.Endpoint && Data.Endpoint.User && Data.Endpoint.User.UserId );
+const consentType = 'statistics' + ( isAnonymous ? '-anonymous' : '' );
```
jazzsequence, line 829: This seems to imply an either/or relationship with statistics cookies, e.g. you can't be both looking at general statistics _and_ anonymous statistics. Is this accurate and is that intentional?

jazzsequence (commented):
a couple comments/questions. overall looks pretty good

inc/namespace.php
roborourke: Ah sorry the intention is for this to work as before but requires you to filter the value elsewhere so it's not directly dependent on the Altis cookie consent plugin. It would work with any implementation of the WP Consent API this way. Maybe I'm overthinking that but I want to keep it somewhat capable of working outside of the Altis context.

src/analytics.js
roborourke: Hm, so I opted to have this switch in place because if the endpoint `UserId` is set that allows the site to link a logged in users endpoints together and to identify them. Without that there's no identifying data tracked by default hence the `isAnonymous` switch. Maybe we need to allow the category to be filtered on the backend instead to provide control and avoid the need for this logic.

roborourke: I changed it so the default is just `statistics` to keep things simple. You know this stuff better regarding the categories, what do you think would make sense here?

I'm a bit confused what you mean by looking at both non-anonymous and anonymous statistics. From a backend / dashboard perspective you certainly could, but this is just about checking for a suitable category based on the data we're tracking.

Thinking about this a bit further are you suggesting we could do some tracking of data if `statistics-anonymous` is allowed but remove / redact some info until `statistics` is consented to as well?

inc/namespace.php
jazzsequence: If you want it to work outside of the Altis context, you could use the WP-Consent-API filter for this that we wrap around, which is `wp_consent_cookie_prefix`. Then it could work either way.

src/analytics.js
jazzsequence: I guess my thought is that `statistics`/`statistics-anonymous` aren't mutually exclusive. I would assume the default to be `statistics-anonymous` -- we're collecting data, but it's not user-identifiable -- and as such, that's allowed through our consent plugin by default. But if we're collecting data that's for the purpose of analytics that _does_ have user-identifying information, then we need permission. 

The check here makes sense, but the two categories are separate. A site can be allowing both anonymous statistics and non-anonymous statistics, or neither and if, in our context, it's always an either/or thing, this is fine.

inc/namespace.php
```diff
@@ -221,20 +221,42 @@ function enqueue_scripts() {
...
+	 * @param string|null $cookie_prefix The consent cookie prefix or false to skip consent check.
+	 */
+	$consent_cookie_prefix = apply_filters( 'altis.analytics.consent_cookie_prefix', null );
+
+	/**
+	 * Filters the consent category required to run analytics.
+	 *
+	 * @param string $consent_category The consent category to allow analytics to run with.
+	 */
+	$consent_category = apply_filters( 'altis.analytics.consent_category', 'statistics' );
```
jazzsequence, line 239: I'm happy with this being filterable externally. And maybe, as you say, we just call it statistics whether it's user-identifying information or not. By default, in the consent plugin, I'm assuming that anonymous statistics are always allowed through, and am bypassing requesting consent for those purposes, so, even if the data itself isn't identifying, it may be better to just ask by default. 

However, I would expect that, if we are doing some anonymous or non-personally-identifiable data tracking, that there would also be something adding a filter in here or modifying this filter to use the anonymous-statistics category, instead. See also https://github.com/humanmade/aws-analytics/pull/131#discussion_r546012126

inc/namespace.php
roborourke: Updated to use this 👍

src/analytics.js
roborourke: Thanks for the feedback and insight - I've updated it so that anon consent allows tracking but until full stats consent is given the `User` property of the endpoint data is stripped off.

inc/namespace.php
roborourke: Just went for hardcoding these in the end. Unlikely they'll change and we're only integrating with the WP consent API here.

src/analytics.js
```diff
@@ -35,6 +36,12 @@ if ( ! Config.PinpointId || ! Config.CognitoId ) {
 	/* eslint-enable quotes */
 }
 
+/**
+ * Get consent types.
+ */
+let hasAnonConsent = Consent.CookiePrefix && document.cookie.match( `${ Consent.CookiePrefix }_statistics-anonymous=allow` );
```
joehoyle, line 42: Is there no JS API in our consent API to check if someone has already consented?

roborourke: There is yeah but this script is loaded super early hence this approach.

joehoyle: Ahh got you! Might be good to add a some notes about that in the doc comment

roborourke: Added, thanks @joehoyle

rmccue (approved):

PR #143, opened 2021-01-21
[60] contributors can assign audiences

jazzsequence:
Previously, if logged in as a contributor, you could add a personalization block to a post, but you could not assign audiences to it. This was because the access that was being requested from the API for the audience post type was `edit` and contributors only have `view` permissions, therefore the API request returned a 403 forbidden error.

this PR updates a couple things that deal both with the functional changes as well as cosmetic limitations of the contributor user role:

* the `getPosts`, `getPost` and `getCurrentPost` functions now take an object of API query arguments so we can pass in `status` and `context` (and any other API query arguments we may need later) more easily when we're dealing with contributors. 
* we check for errors in `sortPosts` which may bubble up if posts were not able to be retrieved from the API, resulting in console errors
* the Audience list now checks whether the current user has the ability to create audiences, and passes that into the `getPosts` wrapper
* the up and down arrows in the Audience list are disabled if a user is not able to edit audiences
* the status toggle is removed if a user is not able to edit audiences

Screen Shot 2021-01-27 at 1 25 56 PM

fixes humanmade/experiments#60

### DIFF
--- a/inc/audiences/namespace.php
+++ b/inc/audiences/namespace.php
@@ -96,6 +96,7 @@ function register_post_type() {
 function maybe_grant_caps( $allcaps ) {
 	$cap_map = [
 		'edit_audiences' => 'edit_pages',
+		'read_audiences' => 'read',
 		'edit_others_audiences' => 'edit_others_pages',
 		'publish_audiences' => 'publish_pages',
 		'read_private_audiences' => 'read_private_pages',
--- a/inc/audiences/namespace.php
+++ b/inc/audiences/namespace.php
@@ -96,7 +96,6 @@ function register_post_type() {
 function maybe_grant_caps( $allcaps ) {
 	$cap_map = [
 		'edit_audiences' => 'edit_pages',
-		'read_audiences' => 'read',
 		'edit_others_audiences' => 'edit_others_pages',
 		'publish_audiences' => 'publish_pages',
 		'read_private_audiences' => 'read_private_pages',
--- a/inc/audiences/namespace.php
+++ b/inc/audiences/namespace.php
@@ -399,7 +399,7 @@ function admin_enqueue_scripts() {
 		'altis-analytics-audience-ui',
 		plugins_url( 'src/audiences/index.css', dirname( __FILE__, 2 ) ),
 		[ 'wp-components' ],
-		'2020-03-19-1'
+		'2021-01-22'
 	);
 }
 
--- a/src/audiences/ui/components/list-row.js
+++ b/src/audiences/ui/components/list-row.js
@@ -140,11 +140,17 @@ const ListRow = props => {
 				</div>
 			</td>
 			<td>
-				<StatusToggle
-					disabled={ ! canEdit }
-					status={ post.status }
-					onChange={ onStatusChange }
-				/>
+				{ canEdit && (
+					<StatusToggle
+						status={ post.status }
+						onChange={ onStatusChange }
+					/>
+				) }
+				{ ! canEdit && post.status === 'publish' && (
+					<Fragment>
+						<span className="active">{ __( 'Active', 'altis-analytics' ) }</span>
+					</Fragment>
+				) }
 			</td>
 			<td>
 				<Estimate audience={ post.audience } horizontal />
--- a/src/audiences/ui/components/list-row.js
+++ b/src/audiences/ui/components/list-row.js
@@ -146,10 +146,8 @@ const ListRow = props => {
 						onChange={ onStatusChange }
 					/>
 				) }
-				{ ! canEdit && post.status === 'publish' && (
-					<Fragment>
-						<span className="active">{ __( 'Active', 'altis-analytics' ) }</span>
-					</Fragment>
+				{ ! canEdit && (
+					<span className="active">{ post.status === 'publish' ? __( 'Active', 'altis-analytics' ) : __( 'Inactive', 'altis-analytics' ) }</span>
 				) }
 			</td>
 			<td>
--- a/src/audiences/ui/data/reducer.js
+++ b/src/audiences/ui/data/reducer.js
@@ -8,6 +8,9 @@ import { unionBy } from 'lodash';
  * @returns {number} Sort result value.
  */
 const sortPosts = ( a, b ) => {
+	if ( a.error || b.error ) {
+		return 0;
+	}
 	if ( a.menu_order < b.menu_order ) {
 		return -1;
 	}
--- a/src/audiences/ui/data/reducer.js
+++ b/src/audiences/ui/data/reducer.js
@@ -8,6 +8,7 @@ import { unionBy } from 'lodash';
  * @returns {number} Sort result value.
  */
 const sortPosts = ( a, b ) => {
+	// Check for errored posts.
 	if ( a.error || b.error ) {
 		return 0;
 	}
--- a/src/audiences/ui/list.js
+++ b/src/audiences/ui/list.js
@@ -316,11 +316,11 @@ const applyWithSelect = withSelect( select => {
 		getIsLoading,
 		getPagination,
 	} = select( 'audience' );
-	const posts = getPosts();
+	const canCreate = select( 'core' ).canUser( 'create', 'audiences' );
 	const loading = getIsLoading();
 	const pagination = getPagination();
+	const posts = canCreate ? getPosts() : getPosts( { context: 'view', status: 'publish' } );
 
-	const canCreate = select( 'core' ).canUser( 'create', 'audiences' );
 
 	return {
 		canCreate,
--- a/src/audiences/ui/list.js
+++ b/src/audiences/ui/list.js
@@ -177,7 +177,7 @@ class List extends Component {
 		const { page, search } = this.state;
 		this.props.onGetPosts( {
 			page: page + 1,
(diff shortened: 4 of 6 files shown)


jazzsequence:
@roborourke let me know if you have any ideas off the top of your head about the bug above. I suspect the permissions may need to be adjusted for contributors for the `/analytics/v1/xbs` endpoint, but I haven't dug into it yet.

inc/audiences/namespace.php
```diff
@@ -96,6 +96,7 @@ function register_post_type() {
 function maybe_grant_caps( $allcaps ) {
 	$cap_map = [
 		'edit_audiences' => 'edit_pages',
+		'read_audiences' => 'read',
```
jazzsequence, line 99: I'm not sure if this is necessary or if it's implicit. However, as the post type is not public, perhaps it actually is necessary.

src/audiences/ui/data/reducer.js
```diff
@@ -8,6 +8,9 @@ import { unionBy } from 'lodash';
  * @returns {number} Sort result value.
  */
 const sortPosts = ( a, b ) => {
+	if ( a.error || b.error ) {
```
jazzsequence, line 11: This is possibly an unrelated fix, however it could be related to the bug that is happening now.

src/audiences/ui/list.js
```diff
@@ -103,13 +103,19 @@ class List extends Component {
 	 */
 	onSearch = event => {
 		const value = event.target.value;
+		const canCreate = wp.data.select( 'core' ).canUser( 'create', 'audiences' );
```
jazzsequence, line 106: I wasn't sure if there was a better/easier way to get `wp.data.select` but just using `select` here threw an error as it was not defined.

roborourke: Use `this.props.canCreate()` here

src/audiences/ui/data/reducer.js
roborourke: We probably shouldn’t be adding posts that are actually errors to the list in the first place. I’d filter those out in the `getPosts()` function and maybe store or set an error in the redux store instead, then we could decide what to do with it / how to display it later

inc/audiences/namespace.php
roborourke: Best way to find out is to try removing it and see if everything still works

src/audiences/ui/data/reducer.js
jazzsequence: This isn't even in relation to posts. This appears to be any link in the element. It's not specific enough and it's erroring before the list is even displayed, hence at least filtering out the errored elements.

Case in point: I believe this causes an error in the "Select Audiences" link and the variant link in the block itself.

roborourke: This sort posts function is just used when adding posts or updating posts in the redux store, it shouldn’t have any relation to the lower level components - I’m saying that errors should not be added to the list of posts in the first place. You could check the response in the getPosts function or in the reducer itself to make sure of that

jazzsequence: ¯\_(ツ)_/¯ I will check it out but it seems to be affecting more than just the post list.

roborourke: The posts are read from the list in the store so if we have an error stored instead of a post it would appear that way. I think in this instance it can be addressed at a lower level if that makes sense.

src/audiences/ui/list.js
```diff
@@ -103,13 +103,19 @@ class List extends Component {
 	 */
 	onSearch = event => {
 		const value = event.target.value;
+		const canCreate = this.props.canCreate();
```
roborourke, line 106: Not sure how this bit was working but `this.props.canCreate` is the value we passed through in the `applyWithSelect` higher order component. Here it's being called like a function when the value will be a boolean.

I'm suggesting removing this here and just checking `this.props.canCreate` below

```suggestion
```

src/audiences/ui/list.js
```diff
@@ -103,13 +103,19 @@ class List extends Component {
...
 			search: value,
 		} );
 		// Query posts by the search term if we don't have an existing request.
 		if ( ! this.props.loading ) {
-			this.props.onGetPosts( this.state.page, value );
+			this.props.onGetPosts( {
+				page: this.state.page,
+				search: value,
+				context: canCreate ? 'edit' : 'view',
+				status: canCreate ? 'publish,draft' : 'publish',
```
roborourke, line 117:
```suggestion
				context: this.props.canCreate ? 'edit' : 'view',
				status: this.props.canCreate ? 'publish,draft' : 'publish',
```

src/audiences/ui/list.js
```diff
@@ -169,7 +175,10 @@ class List extends Component {
 	 */
 	onNextPage = () => {
 		const { page, search } = this.state;
-		this.props.onGetPosts( page + 1, search );
+		this.props.onGetPosts( {
+			page: page + 1,
+			search: search,
```
roborourke, line 180: Presumably we need the same context and status checks parameters here
```suggestion
			search: search,
			context: this.props.canCreate ? 'edit' : 'view',
			status: this.props.canCreate ? 'publish,draft' : 'publish',
```

src/audiences/ui/list.js
```diff
@@ -316,11 +325,13 @@ const applyWithSelect = withSelect( select => {
...
-	const posts = getPosts();
+	const canCreate = select( 'core' ).canUser( 'create', 'audiences' );
 	const loading = getIsLoading();
 	const pagination = getPagination();
-
-	const canCreate = select( 'core' ).canUser( 'create', 'audiences' );
+	const posts = canCreate ? getPosts() : getPosts( {
+		context: 'view',
+		status: 'publish',
+	} );
```
roborourke, line 334: Ternary conditions tend to work best for shorter values, sometimes they have poor readability. It's a judgment call ultimately but for this logic I would probably write it like this in keeping with how you've done it elsewhere in this file.

```suggestion
	const queryArgs = {
		context: canCreate ? 'edit' : 'view',
		status: canCreate ? 'publish,draft' : 'publish',
	};
	const posts = getPosts( queryArgs );
```

What you could do is write a small helper function to get these params like:

```js
const getQueryArgs = ( queryArgs = {}, shouldOverride = false, overrides = {} ) => {
  return Object.assign(
    {
      context: 'view',
      status: 'publish',
    },
    queryArgs,
    shouldOverride ? overrides : {}
  );
}
```

Then use it like

```js
const posts = getPosts( getQueryArgs( {}, canCreate, {
  context: 'edit',
  status: 'publish,draft'
} ) );
```

Something like that anyway. It's not a blocker for me as the use case is limited to just this file so don't feel like you have to do that.

src/audiences/ui/select.js
```diff
@@ -163,8 +163,12 @@ class Select extends Component {
 const applyWithSelect = withSelect( ( select, props ) => {
 	let audiencePost = null;
 
+	// Check if the current user can create audiences. If so, they should have create permissions sent to getPost.
+	const canCreate = select( 'core' ).canUser( 'create', 'audiences' ),
+		queryArgs = canCreate ? { context: 'edit' } : {};
```
roborourke, line 168: I think the compiler might take care of optimisations like this now but it's not necessary to assign multiple values from a single `const` declaration. It causes problems with alignment anyway if spaces aren't used. It's usually work ok with `let` or `var` if your tab width was 4 spaces but here we're not gaining much.

```suggestion
	const canCreate = select( 'core' ).canUser( 'create', 'audiences' );
	const queryArgs = canCreate ? { context: 'edit' } : {};
```

roborourke (changes requested):
Nice one, this is basically there. There are few small code style tweaks and 2 bugs where `canCreate` was being called like a function and the status & context args were missing from the `onNextPage` query args.

Good work pushing through those walls 👍

src/audiences/ui/components/list-row.js
```diff
@@ -140,11 +140,17 @@ const ListRow = props => {
...
+				{ canEdit && (
+					<StatusToggle
+						status={ post.status }
+						onChange={ onStatusChange }
+					/>
+				) }
+				{ ! canEdit && post.status === 'publish' && (
+					<Fragment>
+						<span className="active">{ __( 'Active', 'altis-analytics' ) }</span>
+					</Fragment>
```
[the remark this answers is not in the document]
jazzsequence: The opposite status isn't displayed for Contributors because they don't have access to draft posts they didn't create, and they don't have access to create audiences so I left it out.

src/audiences/ui/list.js
jazzsequence: I think the suggested change probably looks cleaner. I don't think we need a helper function at this point, but I've been wondering if something like it might be useful as I was going through these changes. I think it matters if this is something we come back to in the future and start using the query args more.

src/audiences/ui/select.js
jazzsequence: `¯\_(ツ)_/¯`

jazzsequence:
This should be good for re-review

roborourke (approved):
If we’re happy it’s all still working for editors, authors and admin then let’s goooo 🚀

good work Chris

PR #144, opened 2021-01-27
Prevent converting rule value input type when removing rules

roborourke:
Fixes #95

### DIFF
--- a/src/audiences/ui/components/rule.js
+++ b/src/audiences/ui/components/rule.js
@@ -57,7 +57,6 @@ const RuleInput = props => {
 	const {
 		disabled,
 		currentField,
-		key,
 		name,
 		operator,
 		value,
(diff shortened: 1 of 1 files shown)


src/audiences/ui/components/rule.js
```diff
@@ -118,6 +126,10 @@ const RuleInput = props => {
 										name={ name }
 										type="text"
 										value={ value }
+										onBlur={ () => {
+											// If the value is empty or in the field data convert back to dropdown on blur.
+											setShowDropdown( value === '' || data.indexOf( value ) >= 0 );
```
jazzsequence, line 131: Is there no way to prevent this duplication of the code on line 86?

roborourke: Good shout, I've set it as a const and reused it everywhere the logic was duplicated.

jazzsequence (approved):
👍

PR #146, opened 2021-01-27
Ensure audience estimates don't return values over 100%

roborourke:
Fixes https://github.com/humanmade/altis-analytics/issues/88

### DIFF
--- a/inc/audiences/namespace.php
+++ b/inc/audiences/namespace.php
@@ -471,12 +471,13 @@ function get_estimate( array $audience ) : ?array {
 		return $cache;
 	}
 
+	$unique_count = get_unique_endpoint_count();
 	$result = Utils\query( $query );
 
 	if ( ! $result ) {
 		return [
 			'count' => 0,
-			'total' => get_unique_endpoint_count(),
+			'total' => $unique_count,
 		];
 	}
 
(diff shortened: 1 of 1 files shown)


inc/audiences/namespace.php
```diff
@@ -547,13 +558,14 @@ function get_unique_endpoint_count() : ?int {
 	];
 
 	$cache = wp_cache_get( 'total-uniques', 'altis-audiences' );
-	if ( $cache ) {
+	if ( $cache && ! $force_update ) {
```
jazzsequence, line 561: If the issue has to do with cached values, and we're passing `$force_update` to bypass the cache, should we not cache the new, updated values? Is the cache set elsewhere?

roborourke: Yeah further down in the code. Before this change if the `$cache` value was truthy we'd just return it early here. If the function continues on the query is re-run and then the cache is set at the end of the function before returning the value.

jazzsequence: I figured, since I didn't see the cache set in this code, I just wanted to make sure that was the case.

jazzsequence (approved):
👍

inc/audiences/namespace.php
roborourke: FWIW I regularly use the little blue up/down arrows in the code review view to look at an entire function, or sometimes a while file to give myself context.

jazzsequence: Yeah, I did that, but I still didn't see where the cache was being set in this path. Am I missing it?

roborourke: https://github.com/humanmade/aws-analytics/pull/146/files#diff-2667fc93072cbbc8eb88479fb4b4e173e128dc672084113b9c5c297e7c92db67R574

jazzsequence: I missed it. See it now. 👍

PR #147, opened 2021-01-27
Warn on navigation if audience has edits

roborourke:
Fixes https://github.com/humanmade/altis-analytics/issues/96

### DIFF
--- a/src/audiences/ui/edit.js
+++ b/src/audiences/ui/edit.js
@@ -49,12 +49,31 @@ const StyledEdit = styled.div`
  */
 class Edit extends Component {
 	state = {
-		notice: null,
 		error: null,
+		hasEdits: false,
+		notice: null,
 	};
 
 	static getDerivedStateFromError( error ) {
 		return { error };
+	}
+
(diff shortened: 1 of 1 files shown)


jazzsequence (commented):
this is fine if you are creating an audience from the Audiences page, but if you are creating/editing an audience from the modal inside a post, you never actually "leave the page" so it's possible to make a change, navigate away (e.g. close the modal or click the Back to Audiences button) and then lose the change from within a post.

Not sure if that's within the scope of this ticket, but it seems like it could/should apply to both things. Possibly this issue could be spun up as a separate ticket so we can push this specific change (because it's at least better than before) if the other thing would be a lot of additional work.

roborourke:
Hmm yeah. It’s harder to bubble that up and closing the modal is a more intentional thing than accidentally navigating. Let’s handle that separately, this is addressing the navigation issue rather than the modal in terms of scope I think

jazzsequence (approved):
Created a new issue for this happening within the modal #152

PR #153, opened 2021-01-28
Update cap required to view estimates

### LINKED ISSUE #151: Audience Estimates are not displayed for contributors
With humanmade/experiments#60, Contributors can now select audiences. However the audience estimates are not visible for contributors.

This is because Contributors get a 403 Forbidden error on the REST API request to `https://altis-dev.altis.dev/wp-json/analytics/v1/audiences/estimate?audience=%7B%22include%22%3A%22all%22%2C%22groups%22%3A%5B%7B%22include%22%3A%22any%22%2C%22rules%22%3A%5B%7B%22field%22%3A%22endpoint.Location.Country%22%2C%22operator%22%3A%22%3D%22%2C%22value%22%3A%22JP%22%7D%5D%7D%5D%7D&_locale=user`

jazzsequence:
updates the capability required to view estimates.

Resolves an issue where audience estimates are not displayed if logged in as a Contributor

Before:
screenshot of before state

After:
Screen Shot 2021-01-28 at 3 41 25 PM

Fixes #151

### DIFF
--- a/inc/audiences/rest_api/namespace.php
+++ b/inc/audiences/rest_api/namespace.php
@@ -82,7 +82,7 @@ function init() {
 		[
 			'methods' => WP_REST_Server::READABLE,
 			'callback' => __NAMESPACE__ . '\\handle_estimate_request',
-			'permission_callback' => __NAMESPACE__ . '\\check_edit_permission',
+			'permission_callback' => __NAMESPACE__ . '\\check_read_permission',
 			'args' => [
 				'audience' => [
 					'description' => __( 'A URL encoded audience configuration JSON string', 'altis-analytics' ),
--- a/inc/audiences/rest_api/namespace.php
+++ b/inc/audiences/rest_api/namespace.php
@@ -268,3 +268,13 @@ function check_edit_permission() : bool {
 	$type = get_post_type_object( Audiences\POST_TYPE );
 	return current_user_can( $type->cap->edit_posts );
 }
+
+/**
+ * Check user can view audience posts.
+ *
+ * @return bool
+ */
+function check_read_permission() : bool {
+	$type = get_post_type_object( Audiences\POST_TYPE );
+	return current_user_can( $type->cap->read );
+}


inc/audiences/rest_api/namespace.php
```diff
@@ -268,3 +268,13 @@ function check_edit_permission() : bool {
...
 }
+
+/**
+ * Check user can view audience posts.
+ *
+ * @return bool
+ */
+function check_read_permission() : bool {
+	$type = get_post_type_object( Audiences\POST_TYPE );
+	return current_user_can( $type->cap->read );
```
roborourke, line 279: Let’s use the primitive `edit_posts` capability here instead. I don’t think we want the general public to be able to see what kind of traffic the site is getting. Contributors and up will be able to see them.

roborourke (changes requested):

inc/audiences/rest_api/namespace.php
rmccue: If I have access to work with pages on a site, but not blog posts (i.e. `edit_pages` but not `edit_posts`), presumably I should have access to audiences as well for the same reasoning? Seems to me like using the read cap for audiences is the better path, and then mapping the meta cap under the hood if we want to default it.

roborourke: I'd rather use a custom capability than default to read in that case, then as you say use `map_meta_cap`. My mistake there though, I thought this was the primitive. Will need to update this in the same for viewing XB analytics data in that case too. I'll make a PR for that on the experiments repo.

rmccue: To be clear, when I say read, I mean `$type->cap->read`, not the primitive `'read'`. That can be set to `read_audiences` in the post type registration, then `map_meta_cap` can map that to whatever under the hood

roborourke: Audience posts themselves are public, this is specifically for the analytics data related to audiences

rmccue: 🤔

roborourke: I was wrong actually, this permission check is fine and doesn't need the same permissions as the xb analytics. I think I got the wrong end of the stick there so we should probably use the same check that we have here in the experiments plugin - if you can read audiences which are not actually public (was getting confused because we output those programmatically on the front end) then you should be able to read the analytics data for XBs.

I'll make an issue to correct it on the experiments repo to use the audience read permission..

roborourke (approved):

PR #160, opened 2021-02-02
Add posts controller extension for audience REST API permissions

roborourke:
REST API endpoint permissions are a little different and require extending and modifying the default posts controller class permission checks. This update ensures audiences are not public even if their post status is `publish`, but are still available through the REST API for authenticated requests.

### DIFF
--- a/inc/audiences/rest_api/class-posts-controller.php
+++ b/inc/audiences/rest_api/class-posts-controller.php
@@ -1,0 +1,53 @@
+<?php
+/**
+ * Audience REST API posts controller.
+ *
+ * @package altis/aws-analytics
+ */
+
+namespace Altis\Analytics\Audiences\REST_API;
+
+use WP_Error;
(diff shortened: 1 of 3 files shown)


roborourke:
Checked and contributors can still read audiences ok so there's no regression.

inc/audiences/rest_api/class-posts-controller.php
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	/**
+	 * Checks if a given request has access to read posts.
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 * @return true|WP_Error True if the request has read access, WP_Error object otherwise.
+	 */
+	public function get_items_permissions_check( $request ) {
+
```
rmccue, line 39:
```suggestion
```

rmccue (approved):

PR #164, opened 2021-02-09
Add persistent referer and UTM parameter tracking

### LINKED ISSUE #58: Persist UTM attributes and referrer
**User Story:**
As a marketer, I want to personalize content based on the initial referrer to the website so that I can understand the conversion of my different marketing campaigns (social / email /etc.)

**Requirement:**
All the UTM parameters and the referrer selection should list the audience that came from the very first campaign they were referred by. Basically, the initial referrer should be sticky. Eg: A website visitor who visited through a social media campaign should always be listed under the UTM attribute = social.    

**Acceptance Criteria:**
Given: A user visits the site from a campaign or clicks through a site
When: They visit another page with personalized content 
Then: They should see the personalized content based on the initial referrer they came from

roborourke:
Fixes #58

### DIFF
--- a/src/audiences/ui/components/rule.js
+++ b/src/audiences/ui/components/rule.js
@@ -203,6 +203,9 @@ export default function Rule( props ) {
 	const fields = useSelect( select => select( 'audience' ).getFields(), [] );
 	const currentField = fields.find( fieldData => fieldData.name === field ) || {};
 
+	const persistentFields = fields.filter( fieldData => fieldData.name.match( /^endpoint\./ ) );
+	const pointInTimeFields = fields.filter( fieldData => ! fieldData.name.match( /^endpoint\./ ) );
+
 	return (
 		<StyledRule className="audience-editor__rule">
 			<select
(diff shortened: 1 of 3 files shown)


rmccue (approved):
Makes sense, but maybe there's something better for the microcopy than "Initial Referrer", "Any Referrer", "Current Referrer"; @PriyaBhatia0210 any ideas?

roborourke:
@rmccue yeah I wasn't completely sure on those while writing them so glad you brought it up. I forgot as well that there's optional description text for the field so we can further clarify what it means in longer form.

roborourke:
Could at least change "Any Referrer" to just "Referrer" and add descriptions of what each means eg.

- "Initial Referrer": The first external website a visitor arrived from
- "Referrer": Any external website a visitor has ever arrived from
- "Current Referrer": The previous URL a visitor came from for the current page view only

I was also thinking I could group those options in the field dropdown up into "Persistent" and "Point in time" `<optgroup>`s based on whether it's endpoint data or not. What do you reckon?

src/audiences/ui/components/rule.js
```diff
@@ -221,15 +224,26 @@ export default function Rule( props ) {
...
+					{ persistentFields.map( fieldData => (
+						<option
+							key={ fieldData.name }
+							value={ fieldData.name }
+						>
+							{ fieldData.label }
+						</option>
+					) ) }
+				</optgroup>
+				<optgroup label={ __( 'Point in time data', 'altis-analytics' ) }>
```
rmccue, line 237: `Data` should be consistently capitalised here; also added some hyphens for fun

```suggestion
				<optgroup label={ __( 'Point-in-time Data', 'altis-analytics' ) }>
```

rmccue (approved):
Approved with above caveat

PR #166, opened 2021-02-22
Add shadow post type for XBS

roborourke:
Fixes https://github.com/humanmade/altis-analytics/issues/132

After talking through the solution with @rmccue decided against overloading the reusable block post type or making available in the same so this is much simpler than originally thought.

Tested with nested block and multiple blocks on page. Resetting the block's analytics data e.g. `clientId` will create a new XB post, it'll be treated a new block.

### DIFF
--- a/inc/blocks/namespace.php
+++ b/inc/blocks/namespace.php
@@ -9,9 +9,13 @@ namespace Altis\Analytics\Blocks;
 
 use Altis\Analytics\Audiences;
 use Altis\Analytics\Utils;
+use WP_Post;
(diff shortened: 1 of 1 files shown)


jazzsequence (approved):
Works as advertised. 

I noticed that the xbs endpoint/post type doesn't include any actual data from the experience block (e.g. component props/variants) stored as bits of post meta which I would expect. But as this feature is initially focused on analytics, and that part hasn't been hooked up yet, this looks fine as a first pass.

PR #167, opened 2021-02-23
Add title field to XBs

roborourke:
Fixes https://github.com/humanmade/altis-analytics/issues/128

Opted to show the title in place of the default "Personalized Content" heading too. This will be compatible with #166 as well and has matching placeholder post titles built from the current post title and the XB instance number on the page.

Screenshot 2021-02-23 at 11 39 43

### DIFF
--- a/inc/blocks/personalization/block.json
+++ b/inc/blocks/personalization/block.json
@@ -12,6 +12,9 @@
             "clientId": {
                 "type": "string",
                 "required": true
+            },
+            "title": {
+                "type": "string"
             }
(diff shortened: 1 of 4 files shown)


jazzsequence (commented):
~This looks okay in theory, but I am unable to test. When I run `npm run build` or `npm run start` I get an error in the build.~

Derp, needed to run an `npm install`

jazzsequence (approved):
💥 Works like it says! (after my brain fart)

PR #173, opened 2021-03-12
Add XB analytics single block view

roborourke:
- This adds the single page for XB analytics
- Removes the analytics output from the XB sidebar in favour of a popup modal
- Improves the audience and XB rest API routes
- Move the data store components for XBs and audiences to individually enqueueable files

Fixes https://github.com/humanmade/altis-analytics/issues/138

Screenshot 2021-03-12 at 14 20 13

Screenshot 2021-03-12 at 14 20 27

altis-dev altis dev_wp-admin_admin php_page=xb-analytics clientId=2a7d3480-e525-4fc0-b27d-66d677dd3008


jazzsequence:
I have some design feedback before digging into the code changes.

1. There should be some variation between the block variants. A line or row shading to just distinguish where one thing ends and the next thing begins.
2. I am wondering what it would look like if the values in the three main cards at the top were centered. Block views and Conversion rate seem okay, but Personalization Coverage looks weird. It might be worth seeing what it looks like if you center the big value and then put the smaller (% change) value offset next to the bigger one in the cards that have that. (This might look awful, but it may be worth experimenting).
3. "Block views" and "Conversion rate" should be "Block Views" and "Conversion Rate" if we're capitalizing "Personalization Coverage"
4. I think that we could align the data points in the block variants better. Maybe it's just as simple as a grid row with 5 columns that are equal size across the row, but the empty space to the right looks weird. I almost want to align them with the 3 cards at the top, but I don't really see how that would work so just justifying the columns might be the best way of handling it.
5. I'd be interested to know what the contrast ratio is between the lift v fallback green and the background color is. It looks like it might not be enough to pass WCAG a11y standards. At the very least, it seems like the background is darker than it needs to be for readability. (It might be worth just running the whole page through an accessibility/color contrast checker to see where things fall.)

roborourke:
1. I don’t know, I think it’s ok personally. I made sure there was a good amount of vertical space between them
2. Don’t really agree on centering those metrics. It’d look weirder and be less scannable. Left aligned is generally the best option. I don’t really want to fiddle with the design much, we can always iterate when we get a designer on board
3. Yep 👍 
4. Don’t fear the white space :) they’re aligned with each other in each variant and it looks much better having them more closely associated than spread out across a big space IMO
5. Yeah I don’t think it’s a pass, it’s not as bad as it might be given the larger size and bold weight but I’ll look at a slightly darker version of each
6. They’re not super clear but also not super useful, just a bit of visual interest and flourish really, I’ll increase the contrast a bit

I can implement some of this but I also don’t want to spend ages noodling with stuff we can iterate on when we’re against time. Don’t let it hold up reviewing the code at any rate!

src/blocks/ui/block.js
```diff
@@ -0,0 +1,167 @@
...
+		const fallback = analytics.audiences.find( audience => audience.id === 0 );
+		personalisedCoverage = 100 - ( ( fallback.unique.views / analytics.unique.views ) * 100 );
+	}
+
+	return (
+		<BlockWrapper className="altis-analytics-block">
+			<h1>{ __( 'Experience Insights', 'altis-analytics' ) }</h1>
+			<h2>
+				{ ( block && block.title.rendered ) || __( 'Loading…', 'altis-analytics' ) }
+				{ block && block.parent && (
```
jazzsequence, line 108: This is returning `0` for me, so my title looks like:
```html
<h2>
    XB Analytics Demo
    0
</h2>
```

Screen Shot 2021-03-15 at 12 12 24 PM

We should fail gracefully so a failed conditional doesn't output in the markup.

src/blocks/ui/components/variants.js
```diff
@@ -0,0 +1,145 @@
...
+
+	ul {
+		list-style: none;
+		display: flex;
+		flex-wrap: wrap;
+		margin: 10px 0 0;
+		padding: 0;
+	}
+
+	li {
```
jazzsequence, line 35: Still feel like an alternating light/dark background color for rows would make it more readable, but not a blocker to get this in and worth waiting until we get feedback about the end result to make design tweaks.

jazzsequence (changes requested):
One issue that looks like a bug (see screenshot) but mostly non-blocking feedback/questions otherwise. 👍

inc/blocks/personalization/components/block-analytics.js
```diff
@@ -38,7 +38,7 @@ const BlockAnalytics = ( { clientId } ) => {
 	return (
 		<div className="altis-experience-block-analytics">
 			<h4>{ __( 'Block Insights', 'altis-experiments' ) }</h4>
-			<p>{ __( 'Statistics shown are for the configured data retention period.', 'altis-experiments' ) }</p>
+			<p>{ __( 'Statistics shown are for the last 7 days.', 'altis-experiments' ) }</p>
```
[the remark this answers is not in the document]
roborourke: Ah yeah missed this in the merger, hmm. I'll tackle all of these at once in a separate PR I think.

inc/namespace.php
```diff
@@ -305,10 +305,10 @@ function delete_old_indexes() {
 	$max_age = (int) apply_filters( 'altis.analytics.max_index_age', 14 );
 
 	// If age has been set out of the 60-day limit, default to 7 days and warn user.
```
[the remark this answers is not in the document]
roborourke: Yeah good shout

src/blocks/ui/block.js
roborourke: Oh derp, should be an expression that returns a bool when using that pattern
```suggestion
				{ block && block.parent > 0 && (
```

inc/namespace.php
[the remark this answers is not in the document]
roborourke:
```suggestion
	// If age has been set out of the 90-day limit, default to 7 days and warn user.
```
Also need to consider bumping the default to 90 days but want to discuss with the team first.

src/blocks/ui/components/variants.js
roborourke: I think with the big stripe further up that it would look odd using a background colour to alternate rows, would be a bit too much noise in the UI. Let's see what Noel thinks and come back to it.

src/blocks/ui/block.js
```diff
@@ -0,0 +1,167 @@
...
+		margin: 0 0 20px;
+	}
+`;
+
+/**
+ * Experience Block Analytics component.
+ *
+ * @param {object} props The component props.
+ * @param {string} props.clientId The block client ID.
+ * @param {boolean} props.showPosts Whether to show posts this XB is shown on.
```
roborourke, line 55:
```suggestion
```

roborourke:
Nice one @jazzsequence, made those small fixes now.

src/blocks/ui/components/variants.js
jazzsequence: That's fair. 👍
I'm not in love with the big blue stripe, either, tbh, but `¯\_(ツ)_/¯`

jazzsequence (approved):

PR #174, opened 2021-03-12
analytics dashboard list view

jazzsequence:
fixes https://github.com/humanmade/altis-analytics/issues/137 

Adds the Experience Insights list page.

Screen Shot 2021-03-19 at 1 15 34 PM


inc/dashboard/namespace.php
```diff
@@ -0,0 +1,436 @@
...
+	$cache = wp_cache_get( $key, 'altis-xbs' );
+
+	if ( $cache ) {
+		return $cache;
+	}
+
+	$result = query( $query );
+
+	// Check for a result before getting aggregated data.
+	$data = ( $result ) ? get_aggregate_data( $result['aggregations']['blocks']['buckets'] ?? [] ) : [];
```
roborourke, line 388: Looks like `get_aggregate_data()` always returns an array so the logic isn't needed here:
```suggestion
	$data = get_aggregate_data( $result['aggregations']['blocks']['buckets'] ?? [] );
```

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,436 @@
...
+ * @param \WP_Query $query The WP_Query object.
+ *
+ * @return \WP_Query The updated WP_Query object.
+ */
+function modify_views_list_query( $query ) {
+	if (
+		// Bail if we arent' in the admin.
+		! is_admin() ||
+		// Bail if we aren't looking at the XB Insights page.
+		$query->get( 'post_type' ) !== 'xb'
```
roborourke, line 412: Use constant for post type name.

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,436 @@
...
+ * @todo Finish this function.
+ *
+ * @param \WP_Query $query The WP_Query object.
+ *
+ * @return \WP_Query The updated WP_Query object.
+ */
+function modify_views_list_query( $query ) {
+	if (
+		// Bail if we arent' in the admin.
+		! is_admin() ||
```
roborourke, line 410: Why only in the admin? This default sorting or sorting capability could be useful in REST API responses and other potential use cases too.

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,436 @@
...
+					'order' => [
+						'_count' => $order,
+					],
+				],
+				'aggs' => [
+					'events' => [
+						'terms' => [
+							'field' => 'event_type.keyword',
+						],
+					],
```
roborourke, line 368:
```suggestion
					],
					'aggs' => [
						'uniques' => [
							'cardinality' => 'endpoint.Id.keyword',
						],
					],
```

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,436 @@
...
+ *
+ * @uses get_views_List()
+ *
+ * @param string $block_id Get a single block's analytics data pulled from the ES query.
+ *
+ * @return array The block's analytics data.
+ */
+function get_block_data( string $block_id ) : array {
+	$data = get_views_list();
+	return $data[ $block_id ];
```
roborourke, line 319: What happens if `$block_id` is not in the array?

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,436 @@
...
+ * @return array The sorted array.
+ */
+function sort_by_conversion_rate( array $list, string $order = 'desc' ) : array {
+	// If an invalid value was passed to $order, default to 'desc'.
+	if ( ! in_array( $order, [ 'asc', 'desc' ], true ) ) {
+		$order = 'desc';
+	}
+
+	$avg_conversion_rate = array_column( $list, 'avg_conversion_rate' );
+	$sort_order = 'SORT_' . strtoupper( $order );
```
roborourke, line 301: The sort flags are constants not strings so I'm not sure this will work. Use something like:
```suggestion
	$sort_order = ( $order === 'desc' ) ? SORT_DESC : SORT_ASC;
```

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,436 @@
...
+
+/**
+ * Calculate average conversion rate.
+ *
+ * @param array $block_data An array of block data.
+ * @param string $block_id A single block clientID.
+ *
+ * @return float The conversion rate (calculated by conversions / views).
+ */
+function calculate_average_conversion_rate( array $block_data = [], string $block_id = '' ) : float {
```
roborourke, line 240: In every instance this function has the `$block_data` passed into it and never the `$block_id`, seems it could be simplified to just accept views and conversions as arguments instead.

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,436 @@
...
+		! is_admin() ||
+		// Bail if we aren't looking at the XB Insights page.
+		$query->get( 'post_type' ) !== 'xb'
+	) {
+		return $query;
+	}
+
+	$order = $query->get( 'order' ) ?: 'desc';
+	$orderby = $query->get( 'orderby' ) ?: 'views';
+	$days = get_days_view();
```
roborourke, line 419: Might be good to support this via a query arg too, eg. `$days = $query->get( 'days' ) ?: get_days_view();`

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,436 @@
...
+		// Bail if we arent' in the admin.
+		! is_admin() ||
+		// Bail if we aren't looking at the XB Insights page.
+		$query->get( 'post_type' ) !== 'xb'
+	) {
+		return $query;
+	}
+
+	$order = $query->get( 'order' ) ?: 'desc';
+	$orderby = $query->get( 'orderby' ) ?: 'views';
```
roborourke, line 418: Seems we should check at this point whether we're sorting by views or conversions and bail if not? It's only those 2 that require the ES query from the looks of it.

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,436 @@
...
+		return $cache;
+	}
+
+	$result = query( $query );
+
+	// Check for a result before getting aggregated data.
+	$data = ( $result ) ? get_aggregate_data( $result['aggregations']['blocks']['buckets'] ?? [] ) : [];
+
+	// Don't cache anything if we didnt' get a result.
+	if ( ! empty( $data ) ) {
```
roborourke, line 391: I learned this recently but getting no data is also a valid response and should be cached. You could check this prior to the `get_aggregate_data()` call. Typically what I've done is to set a lower cache time when there's no data like a minute, then 5 minutes when we do have data.

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,436 @@
...
+ */
+function render_last_modified_author() {
+	global $post;
+	?>
+	<div class="post--last-modified">
+		<span class="post--last-modified-date"><?php echo esc_attr( get_the_modified_date( '', $post ) ); ?></span>
+		<span class="post--last-modified-time"><?php echo esc_attr( get_the_modified_time( '', $post ) ); ?></span>
+	</div>
+	<div class="post--author">
+		<?php echo esc_attr( get_the_author_meta( 'display_name', $post->post_author ) ); ?>
```
roborourke, line 190: Posted a question on the issue but what're your thoughts on this juxtaposition of data? Does it make sense to you that these are combined?

roborourke (changes requested):
Cheers @jazzsequence, looks largely there but there are a few important bugs to fix up like the date range filter.

Seems like an errant `composer.lock` file has been committed too, that needs to be removed.

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,436 @@
...
+ * @param array $columns An array of columns to filter.
+ * @param string $post_type The current post type.
+ *
+ * @return array The filtered columns array.
+ */
+function remove_default_columns( $columns, $post_type ) : array {
+	if ( 'xb' === $post_type ) {
+		unset( $columns['cb'] );
+		unset( $columns['title'] );
+		unset( $columns['altis_publication_checklist_status'] );
```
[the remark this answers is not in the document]
jazzsequence: Since there are no (and will be no) bulk actions for items in this list, there's no reason for them to have checkboxes. But I can see if I can figure out how to remove post type support for this.

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,436 @@
...
+	$columns['conversion'] = 'conversion';
+	return $columns;
+}
+
+/**
+ * Add additional microcopy to column titles.
+ *
+ * Note: This adds an additional span around the column title as there is no current way to filter the column title span itself.
+ *
+ * @see https://github.com/WordPress/WordPress/blob/643888dfc6697877609672087b4d3b4e5bf545ff/wp-admin/includes/class-wp-posts-list-table.php#L1059
```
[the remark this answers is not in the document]
jazzsequence: That makes sense. I was concerned about linking directly to a file in, say, the `master` branch because the line number isn't guaranteed to be the same, and context is important. Theoretically, that could happen in a branch, too, but possibly less often? Anyway, linking to a specific commit at least means that the link/line I'm pointing to wouldn't ever change. That was the thinking anyway. `¯\_(ツ)_/¯`

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,436 @@
...
+ *
+ * @param array $xb_columns An array of columns for the XB post type.
+ *
+ * @return array The filtered array of columns.
+ */
+function add_microcopy_to_column_titles( array $xb_columns ) : array {
+	$microcopy = [
+		'block' => __( 'List of XBs modified in the selected date range', 'altis-analytics' ),
```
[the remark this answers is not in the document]
jazzsequence: This was the microcopy that we had agreed on in the original issue https://github.com/humanmade/altis-analytics/issues/137. We can come back to it again if it seems like it's an inaccurate representation. 

From a technical perspective, it is a list of XBs, with analytics data reflected in the selected date range. We're not actually filtering out stuff that has no data, but I could do. @PriyaBhatia0210 thoughts?

inc/dashboard/namespace.php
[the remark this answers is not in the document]
roborourke: Sorry - should have been more specific, I was referring to the publication checklist. Review comments are generally associated with the last line in what’s shown, bit confusing as we can do multiline comments now.

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,436 @@
...
+ *
+ * This displays the title, but the column is a custom column called "block", so we need to change the orderby value to "title".
+ *
+ * @param array $vars An array of query variables passed in from the filter.
+ *
+ * @return array The updated query variables.
+ */
+function xb_block_column_orderby( $vars ) : array {
+	if ( isset( $vars['orderby'] ) && 'block' === $vars['orderby'] ) {
+		$vars['orderby'] = 'title';
```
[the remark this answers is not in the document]
jazzsequence: TBH I can't remember and I went back and forth a couple times on changing it back at various points. The column heading and the fact that the column name reflected that was one reason, but I'm already filtering the column headings, so probably easiest to change it back. 👍

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,436 @@
...
+	<?php
+}
+
+/**
+ * Render the block name column data.
+ */
+function render_block_column() {
+	global $post;
+	?>
+	<strong><a href="<?php echo esc_url_raw( admin_url( '/admin.php?page=xb-analytics&post=' . $post->ID ) ); ?>"><?php echo esc_attr( $post->post_title ); ?></a></strong>
```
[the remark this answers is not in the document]
jazzsequence: I'd like to chat through how the XB post types are hierarchically linked, I'm not sure I'm clear on that yet.

inc/dashboard/namespace.php
jazzsequence: I saw that in the Blocks namespace but didn't (and still don't, really) understand why you'd want to cache nothing for a minute rather than just recheck next time.

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,441 @@
...
+ * Remove the post quick actions.
+ *
+ * @param array $actions An array of actions that can be done on a single post.
+ * @param object $post The WP_Post object
+ *
+ * @return array The filtered array (empty for xbs).
+ */
+function remove_post_row_actions( $actions, $post ) : array {
+	if ( 'xb' === $post->post_type ) {
+		unset( $actions['edit'] );
```
roborourke, line 65: Might be useful to have a link to edit the post the XB is on? Though I think if we're removing the default title column these aren't shown anyway.

inc/dashboard/namespace.php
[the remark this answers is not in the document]
roborourke: By only including stuff that has data in the selected date range it's already filtering out XBs with no analytics data.

inc/dashboard/namespace.php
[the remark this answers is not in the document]
roborourke: Pretty much just this https://github.com/humanmade/aws-analytics/blob/master/inc/blocks/namespace.php#L112

The parent is the post the XB was created on.

inc/dashboard/namespace.php
jazzsequence: they're are shown anyway -- I think they display by default for post list tables unless you explicitly remove them (like I did). Editing the original post from the list seems like a good idea, but what if the same block is used in multiple places (we don't surface that now, but we've discussed maybe doing it in the future)? Might be easier not to lock that in and then have to change it later

inc/dashboard/namespace.php
roborourke: Performance mainly. It's like asking for data and the service correctly saying "there isn't any" and responding with "I don't believe you". There are cases where there might not be any data and this query gets hammered, especially as each row has callbacks for the column data that use this function. If I'm showing 50 XBs per page and sorting by date or title for example if there are no views then this query will be run 100 times in quick succession, once for each views and conversion rate column.

inc/dashboard/namespace.php
jazzsequence: Based on that convo, and the idea that these might need to be sortable as well, let's split them into separate columns.

inc/dashboard/namespace.php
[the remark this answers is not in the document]
jazzsequence: That makes more sense now. 😅

inc/dashboard/namespace.php
jazzsequence: This throws an error after I merged it in.

```Analytics: elasticsearch query failed: http://elasticsearch:9200/analytics*/_search {"query":{"bool":{"filter":[{"term":{"attributes.blogId.keyword":1}},{"terms":{"event_type.keyword":["experienceView","conversion"]}}]}},"aggs":{"blocks":{"terms":{"field":"attributes.clientId.keyword","size":10000,"order":{"_count":"asc"}},"aggs":{"events":{"terms":{"field":"event_type.keyword"}},"aggs":{"uniques":{"cardinality":"endpoint.Id.keyword"}}}}},"size":0} {"error":{"root_cause":[{"type":"named_object_not_found_exception","reason":"[1:318] unable to parse BaseAggregationBuilder with name [uniques]: parser not found"}],"type":"named_object_not_found_exception","reason":"[1:318] unable to parse BaseAggregationBuilder with name [uniques]: parser not found"},"status":400}```

File: inc/dashboard/namespace.php (hunk 14 above)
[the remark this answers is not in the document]
Comment (jazzsequence) [reply inc/dashboard/namespace.php:111]: That doesn't seem to be accurate to what I have locally. I can see XBs that don't have any data (or all the data is 0). But maybe 0 is different than no data?

File: inc/dashboard/namespace.php (hunk 15 above)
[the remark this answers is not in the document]
Comment (jazzsequence) [reply inc/dashboard/namespace.php:136]: Oh, now I remember. Using the default title column adds a whole bunch of specific stuff that would need to be changed/filtered (e.g. linking to the single XB -- which we don't want to expose -- as well as (now that it's hierarchical) displaying the parent post next to the post title). Changing it allows us to have more control over what's actually displayed.

File: inc/dashboard/namespace.php (hunk 16 above)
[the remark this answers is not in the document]
Comment (jazzsequence) [reply inc/dashboard/namespace.php:193]: After looking at this filter, I think it's better to pass on it here. The `$context` parameter doesn't really help us out as much as I would like it to and the two places we could use it (here, to link to the analytics view and in the quick edit row, to link to the parent post) are two very different destinations. If `$context` gave us actually useful information, like where the link was being used or something, this would be brilliant, but as it is, for both those links it just returns `'display'` which is totally unhelpful.

That said, we probably _should_ still add the filter -- and always link back to the parent post -- and just not use it on this line of code, specifically, so that we never end up on a edit single block page.

File: inc/dashboard/namespace.php (hunk 9 above)
Comment (jazzsequence) [reply inc/dashboard/namespace.php:421]: we want to sort by views by default, so we always want to run the ES query. Otherwise it will sort by block title by default.

File: inc/dashboard/namespace.php (hunk 4 above)
Comment (roborourke) [reply inc/dashboard/namespace.php:362]: Oh whoops, that sub `aggs` should be inside the terms aggregation for events, at the same level as the `field` property. My fault

File: inc/dashboard/namespace.php (hunk 14 above)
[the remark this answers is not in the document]
Comment (roborourke) [reply inc/dashboard/namespace.php]: That would depend on whether they're being sorted by views or conversions, anything else would show all XBs because it's not using `post_name__in` for the query. Ruling that out it's filtering by conversions and views in the query so technically there has to be data if anything is to be returned from ES.

File: inc/blocks/namespace.php [hunk 18]
```diff
@@ -191,6 +220,25 @@ function register_post_type() {
...
+ * @param int $post_id The post ID.
+ *
+ * @return string The updated edit post link.
+ */
+function update_xb_edit_post_link( string $link, int $post_id ) : string {
+	if ( get_post_type( $post_id ) !== POST_TYPE ) {
+		return $link;
+	}
+
+	$parent_id = wp_get_post_parent_id( $post_id );
```
roborourke, line 236: I think this function was just added in WP 5.7, we maybe don't need to worry about back compat in this instance but if this was going to part of a preview release or someone wanted to use it with Altis < 7 it'd break.

inc/blocks/namespace.php
```diff
@@ -191,6 +220,25 @@ function register_post_type() {
...
+	$updated_link = $parent_id ? str_replace( $post_id, $parent_id, $link ) : $link;
```
roborourke, line 237: This should be ok but concerns me a bit for some reason. We could perhaps return early if there's no parent ID, and ideally we don't want to pass back the default edit link either, what might be a reasonable default? Empty string or null? If WP doesn't like that maybe link to the XB list view 🤔 

Besides that I'm thinking `add_query_arg()` might be a more intentional and concrete way to replace the existing post ID value, eg:
```suggestion
	if ( ! $parent_id ) {
		return '';
	}
	$updated_link = add_query_arg( [ 'post' => $parent_id ], $link );
```

inc/blocks/namespace.php
```diff
@@ -180,9 +183,35 @@ function register_post_type() {
...
+				'all_items' => __( 'Insights', 'altis-analytics' ),
+			],
+			'admin_cols' => [
+				'block' => [
+					'title' => __( 'Block', 'altis-analytics' ),
+					'function' => '\\Altis\\Analytics\\Dashboard\\render_block_column',
+				],
+				'views' => [
+					'title' => __( 'Views', 'altis-analytics' ),
+					'function' => '\\Altis\\Analytics\\Dashboard\\render_views',
```
roborourke, line 204: Extended CPTs supports setting the default sort column here so you don't need to do any special handling yourself:
```suggestion
					'function' => '\\Altis\\Analytics\\Dashboard\\render_views',
					'default' => 'DESC',
```
https://github.com/johnbillion/extended-cpts/wiki/Admin-columns#default-sort-column

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,439 @@
...
+			],
+		],
+		'aggs' => [
+			'blocks' => [
+				'terms' => [
+					// Get the block data. This will give us the key for the block, which is stored as the post slug.
+					'field' => 'attributes.clientId.keyword',
+					'size' => 10000, // Use arbitrary large size that is more than we're likely to need.
+					'order' => [
+						'_count' => $order,
```
roborourke, line 354: Ordering by count here is going to sort the results by the number of both views _and_ conversions for each XB so won't necessarily be an accurate reflection of the sorting we want.

This should reference the sub aggregation for uniques depending on the type of event we want to sort by.

In looking into this more I'm thinking about a few things:

- How can we keep the number of different queries to ES to a minimum
- Should we try and collect all the data at once
- If we do collect all the data at once can we reference that single cached source of truth rather than calling `Blocks\get_views()`

Things worth noting here too:

- There is a way we can get ES to calculate the conversion rate for us and return it using a bucket script aggregation
- If we're using PHP to sort the XBs in some cases anyway we can reduce the cache variance if we always sort in PHP

inc/dashboard/namespace.php
jazzsequence: Hmm...that's giving me a different error.
```
Analytics: elasticsearch query failed: http://elasticsearch:9200/analytics*/_search {"query":{"bool":{"filter":[{"term":{"attributes.blogId.keyword":1}},{"terms":{"event_type.keyword":["experienceView","conversion"]}}]}},"aggs":{"blocks":{"terms":{"field":"attributes.clientId.keyword","size":10000,"order":{"_count":"asc"}},"aggs":{"events":{"terms":{"field":"event_type.keyword","aggs":{"uniques":{"cardinality":"endpoint.Id.keyword"}}}}}}},"size":0} {"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"[terms] unknown field [aggs], parser not found"}],"type":"illegal_argument_exception","reason":"[terms] unknown field [aggs], parser not found"},"status":400}
```

inc/dashboard/namespace.php
jazzsequence: Right, gotcha. 👍

inc/blocks/namespace.php
jazzsequence: Docs page says it was added in 3.1. http://developer.wordpress.org/reference/functions/wp_get_post_parent_id/

inc/blocks/namespace.php
jazzsequence: 🤔
I guess an empty string makes sense as it would just keep you on the list view.

jazzsequence:
@roborourke this should be g2g now 🎉

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,456 @@
...
+ *
+ * @param \WP_Query $query The WP_Query object.
+ *
+ * @return \WP_Query The updated WP_Query object.
+ */
+function modify_views_list_query( $query ) {
+	// Bail if we aren't looking at the XB Insights page or if we're ordering by block title.
+	if (
+		$query->get( 'post_type' ) !== Blocks\POST_TYPE ||
+		in_array( $query->get( 'orderby' ), [ 'title', 'author', 'last_modified' ], true )
```
roborourke, line 435: What if I sort by some post meta or ID or any other option that WP supports? When we have a known smaller set of values that we want to process in a specific way it makes more sense to return early based on those rather than a handful of other possible options.

I also find it a bit neater and less potentially confusing to have a bunch of separate `if` statements for returning early quite often. It's more verbose but the logic is easier to see and step through when using xdebug for example. Basically I just try and avoid `||` operators as much as possible!

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,456 @@
...
+
+	foreach ( $buckets as $block ) {
+		$block_id = $block['key'];
+		$views = $block['views']['uniques']['value'] ?? 0;
+		$conversions = $block['conversions']['uniques']['value'] ?? 0;
+		$conversion_rate = $block['conversionRate']['value'] ?? 0;
+
+		$data[ $block_id ]['views'] = $views;
+		$data[ $block_id ]['conversions'] = $conversions;
+		$data[ $block_id ]['avg_conversion_rate'] = $conversion_rate;
```
roborourke, line 276: Just occurred to me but the `avg_` bit isn't really what we're looking at here. We're not taking an average of conversion rates across a bunch of blocks or time ranges so this is just the conversion rate, no averages involved.

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,456 @@
...
+
+/**
+ * Calculate average conversion rate.
+ *
+ * @param int $conversions The number of conversions for a block.
+ * @param int $views The number of views for a block.
+ *
+ * @return float The conversion rate (calculated by conversions / views).
+ */
+function calculate_average_conversion_rate( int $conversions = 0, int $views = 0 ) : float {
```
roborourke, line 248: Can remove this function now.

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,456 @@
...
+function sort_by( array $list, string $order = 'desc', string $sort = 'views' ) : array {
+	// If an invalid value was passed to $order, default to 'desc'.
+	if ( ! in_array( $order, [ 'asc', 'desc' ], true ) ) {
+		$order = 'desc';
+	}
+
+	// If an invalid value was passed to $sort, default to 'views'.
+	if ( ! in_array( $sort, [ 'avg_conversion_rate', 'views' ], true ) ) {
+		$sort = 'views';
+	}
```
roborourke, line 300: I probably wouldn't bother with this validation, it'd error out at the `array_column` bit anyway. Someone using the function might use it incorrectly but not be aware because it sorts things out for them, but potentially makes it behave unexpectedly. I know the available values are in the doc block but I think causing a warning or error is a better thing to do if a function is used wrong.

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,456 @@
...
+						],
+						'aggs' => [
+							'uniques' => [
+								'cardinality' => [
+									'field' => 'endpoint.Id.keyword',
+								],
+							],
+						],
+					],
+					'conversionRate' => [
```
roborourke, line 384: If you prefer not to use camel case here you don't have to :)

roborourke (changes requested):
Nice work @jazzsequence, couple of little things to tidy up and some suggestions. Very very nearly there on this now.

roborourke:
Oh I noticed in the screenshot that Query Monitor is showing a warning or something, guessing that's been resolved or is unrelated to this code?

inc/blocks/namespace.php
```diff
@@ -180,9 +183,42 @@ function register_post_type() {
...
+					'default' => 'DESC',
+				],
+				'conversion' => [
+					'title' => __( 'Avg. Conversion', 'altis-analytics' ),
+					'function' => '\\Altis\\Analytics\\Dashboard\\render_average_conversion_rate',
+					'default' => 'DESC',
```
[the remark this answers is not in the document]
jazzsequence: Doesn't seem to change the behavior -- default is still views -- but I'll it out to not be confusing.

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,456 @@
...
+	if ( Blocks\POST_TYPE !== $post->post_type ) {
+		return $actions;
+	}
+
+	unset( $actions['view'] );
+	unset( $actions['trash'] );
+	unset( $actions['inline hide-if-no-js'] );
+
+	// Change the default edit quick action to link to the originating post the block was created on. This should link to a reusable block if it was created as one.
+	$actions['edit'] = sprintf( wp_kses_post( __( '<a href="%1$s">Edit original post</a> (%2$s)', 'altis-analytics' ) ), get_edit_post_link(), get_the_title( $post->post_parent ) );
```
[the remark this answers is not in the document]
jazzsequence: I guess I was thinking it would make things easier, since I'm using `sprintf` anyway, could probably rearrange this. 👍

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,456 @@
...
+/**
+ * Sort data by conversion rate.
+ *
+ * @param array $list The array of analytics data by block id.
+ * @param string $order The order to sort by. Accepted values are 'asc' or 'desc'.
+ * @param string $sort The parameter to sort by. Accepted values are 'views' and 'avg_conversion_rate'. Default is 'views'.
+ *
+ * @return array The sorted array.
+ */
+function sort_by( array $list, string $order = 'desc', string $sort = 'views' ) : array {
```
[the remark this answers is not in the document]
jazzsequence: All good thoughts. I had considered swapping $order and $sort to put the more "important" stuff in front, but there were several cases where I wasn't using/didn't need the $sort value (because I was using the default) and I _was_ using the order. But if this is made to be more generic, perhaps that just needs to be rethought and it's not a huge deal...

Re validation check -- would using a filterable array of values to check against make sense? E.g. you add a filter which passes the values you want to validate against in this context before using the function, or am I overthinking things?

inc/dashboard/namespace.php
jazzsequence: > What if I sort by some post meta or ID or any other option that WP supports? When we have a known smaller set of values that we want to process in a specific way it makes more sense to return early based on those rather than a handful of other possible options.

Not sure I follow the suggestion here.

inc/dashboard/namespace.php
jazzsequence: Hmm...good point, so we'll need to change the column headers and microcopy as well.

jazzsequence:
> Oh I noticed in the screenshot that Query Monitor is showing a warning or something, guessing that's been resolved or is unrelated to this code?

There was something related to this code that I caught, but the notice I'm getting now is `Undefined property: stdClass::$read` in `extra-packages/aws-analytics/inc/blocks/namespace.php:346` which is actually coming from the caps set on your admin page when I backtrace the issue. https://github.com/humanmade/aws-analytics/blob/master/inc/blocks/namespace.php#L285

inc/dashboard/namespace.php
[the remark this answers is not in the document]
jazzsequence: I'm overthinking things. You've said below that validating those passed values isn't a big deal, so I'll just pull that out when I make this reusable.

inc/dashboard/namespace.php
jazzsequence: I got there in the end.... 😅

inc/dashboard/namespace.php
jazzsequence: I was hanging onto it until I knew for sure the ES query for that was working and then forgot to take it out. 😅

inc/dashboard/namespace.php
jazzsequence: switched to snake case because that's what `bucket_script` uses

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,421 @@
...
+		return $query;
+	}
+
+	$order = $query->get( 'order' ) ?: 'desc';
+	$orderby = $query->get( 'orderby' ) ?: 'views';
+	$days = $query->get( 'days' ) ?: get_days_view();
+	$list = get_views_list( $days );
+
+	// Sort by conversion or views.
+	$list = ( $orderby === 'conversion' ) ? sort_by( $list, $order, 'conversion_rate' ) : Utils\sort_by( $list, $orderby, $order );
```
roborourke, line 411: Just a bit of clean up and fixing bug for sorting by conversion rate:
```suggestion
	if ( $orderby === 'conversion' ) {
		$orderby .= '_rate';
	}
	$list = Utils\sort_by( $list, $orderby, $order );
```

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,421 @@
...
+ * @return \WP_Query The updated WP_Query object.
+ */
+function modify_views_list_query( $query ) {
+	// Bail if we aren't looking at the XB Insights page.
+	if ( $query->get( 'post_type' ) !== Blocks\POST_TYPE ) {
+		return $query;
+	}
+
+	// Bail for queries that aren't ordered by views or conversions (so we don't need to run an ES query).
+	if ( ! in_array( $query->get( 'orderby' ), [ 'views', 'conversions' ], true ) ) {
```
roborourke, line 401: Expected value for sorting by conversions `conversion`:
```suggestion
	if ( ! in_array( $query->get( 'orderby' ), [ 'views', 'conversion' ], true ) ) {
```

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,421 @@
...
+ * Alter the default WP_Query to change the sort order.
+ *
+ * @param \WP_Query $query The WP_Query object.
+ *
+ * @return \WP_Query The updated WP_Query object.
+ */
+function modify_views_list_query( $query ) {
+	// Bail if we aren't looking at the XB Insights page.
+	if ( $query->get( 'post_type' ) !== Blocks\POST_TYPE ) {
+		return $query;
```
roborourke, line 397: This is an action rather than a filter so there's no need to return any value, we're modifying the reference to the query object.
```suggestion
		return;
```

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,421 @@
...
+	}
+
+	// Bail for queries that aren't ordered by views or conversions (so we don't need to run an ES query).
+	if ( ! in_array( $query->get( 'orderby' ), [ 'views', 'conversions' ], true ) ) {
+		return $query;
```
roborourke, line 402:
```suggestion
		return;
```

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,421 @@
...
+	$list = ( $orderby === 'conversion' ) ? sort_by( $list, $order, 'conversion_rate' ) : Utils\sort_by( $list, $orderby, $order );
+
+	// Pluck the client ids out of the list.
+	$client_ids = array_keys( $list );
+
+	// Order by client ID (stored as post slug).
+	$query->set( 'post_name__in', $client_ids );
+	$query->set( 'orderby', 'post_name__in' );
+
+	return $query;
```
roborourke, line 420:
```suggestion
```

inc/utils/namespace.php
```diff
@@ -466,3 +466,26 @@ function find_best_accept_header_match( array $parsed, array $available ) : ?str
...
+ * Sort data by conversion rate.
+ *
+ * @param array $list The array of analytics data by block id.
+ * @param string $orderby The parameter to sort by.
+ * @param string $order The order to sort by. Accepted values are 'asc' or 'desc'.
+ *
+ * @return array The sorted array.
+ */
+function sort_by( array $list, string $orderby, string $order = 'desc' ) : array {
+	// If an invalid value was passed to $order, default to 'desc'.
```
roborourke, line 480:
```suggestion
	$order = strtolower( $order );

	// If an invalid value was passed to $order, default to 'desc'.
```

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,421 @@
...
+}
+
+/**
+ * Alter the default WP_Query to change the sort order.
+ *
+ * @param \WP_Query $query The WP_Query object.
+ *
+ * @return \WP_Query The updated WP_Query object.
+ */
+function modify_views_list_query( $query ) {
```
roborourke, line 394:
```suggestion
 * @param WP_Query $query The WP_Query object.
 * @return void
 */
function modify_views_list_query( WP_Query $query ) {
```

inc/dashboard/namespace.php
```diff
@@ -0,0 +1,421 @@
...
+/**
+ * Dashboard Analytics
+ *
+ * @package aws-analytics
+ */
+
+namespace Altis\Analytics\Dashboard;
+
+use Altis\Analytics\Blocks;
+use Altis\Analytics\Utils;
```
roborourke, line 11: Nicer to use WP_Query and add the type hint to the callback.
```suggestion
use Altis\Analytics\Utils;
use WP_Query;
```

roborourke (approved):
Couple of small details to iron out and 2 bugs with the orderby logic but gonna commit those and get this merged 🚀

PR #179, opened 2021-03-18
WordPress 5.7 compat fixes

roborourke:
Related to https://github.com/humanmade/product-dev/issues/670

### DIFF
--- a/inc/blocks/personalization/edit.css
+++ b/inc/blocks/personalization/edit.css
@@ -44,6 +44,14 @@
 	color: inherit;
 }
 
+.altis-experience-block-header__toolbar .components-button .dashicon {
+	margin-right: 0;
+}
+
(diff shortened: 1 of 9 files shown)


jazzsequence (approved):
Looks good for the most part, just a couple questions:

1) Is this all really from the WP 5.7 update? Or were there other bugfixes mixed in?
2) This:
Screen Shot 2021-03-18 at 12 19 34 PM
While (negative) Infinity% is entertaining, probably we should display something else there. 😂

roborourke:
Oh whoops. Yeah there were a few bug fixes that I only noticed after upgrading so I could split this out.

roborourke:
Hmm, interestingly `-Infinity` is a legit occurrence if there are no views from the current 7 days but some from the 7 days prior. Ultimately it would be like `-1 / 0`. Hm. Maybe the infinity symbol instead... Or just nothing 🤔

PR #181, opened 2021-03-24
Add validation for XB conversion goals

roborourke:
This adds a basic check for the conversion goal by using server side rendering to test the completed piece of content (as far as possible).

This does not currently prevent publishing or pop up any warning when updating or publishing the page. Not sure how best to do that yet and if we absolutely need to. Will wait for feedback on the work so far.

Fixes https://github.com/humanmade/altis-analytics/issues/82

Screenshot 2021-03-25 at 18 56 39
Screenshot 2021-03-25 at 18 57 33

### DIFF
--- a/inc/blocks/namespace.php
+++ b/inc/blocks/namespace.php
@@ -10,6 +10,8 @@ namespace Altis\Analytics\Blocks;
 use Altis\Analytics;
 use Altis\Analytics\Experiments;
 use Altis\Analytics\Utils;
+use Altis\Workflow\PublicationChecklist;
+use Altis\Workflow\PublicationChecklist\Status;
 use WP_Post;
 use WP_Query;
 
@@ -50,6 +52,9 @@ function setup() {
 	// Shim server side render block to allow passing inner blocks as attribute.
 	add_filter( 'render_block_data', __NAMESPACE__ . '\\ssr_inner_blocks_shim' );
 	register_ssr_inner_blocks_shim();
+
+	// Publication checklist integration.
+	add_action( 'altis.publication-checklist.register_prepublish_checks', __NAMESPACE__ . '\\check_conversion_goals' );
 }
 
 /**
@@ -707,7 +712,10 @@ function ssr_inner_blocks_shim( array $block ) : array {
 		return $block;
 	}
 
+	// Populate inner blocks by parsing content.
 	$block['innerBlocks'] = parse_blocks( $block['attrs']['content'] ?? [] );
+
+	// Populate inner content with an array of null placeholders so inner blocks are processed.
 	$block['innerContent'] = array_fill( 0, count( $block['innerBlocks'] ), null );
 
 	return $block;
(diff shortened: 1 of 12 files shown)


inc/blocks/namespace.php
```diff
@@ -691,3 +700,99 @@ function get_views( string $block_id, $args = [] ) {
...
+	// Populate inner content with an array of null placeholders so inner blocks are processed.
+	$block['innerContent'] = array_fill( 0, count( $block['innerBlocks'] ), null );
+
+	return $block;
+}
+
+/**
+ * Registers an inner blocks SSR shim.
+ *
+ * @return void
```
jazzsequence, line 727: If there's no return value, do we need to document it? (I generally omit `@return` if there's no actual return, and save `@return void` for when there _is_ a return, but no actual _data_ is returned.)

inc/blocks/namespace.php
```diff
@@ -691,3 +700,99 @@ function get_views( string $block_id, $args = [] ) {
...
+		'render_callback' => function ( array $attributes, ?string $inner_content ) : string {
+			return sprintf( '<div>%s</div>', $inner_content );
+		},
+	] );
+}
+
+/**
+ * Register conversion goal validation check.
+ *
+ * @return void
```
jazzsequence, line 745: This return value is incorrect. If the post type doesn't support Gutenberg, we return early with a `bool`. Otherwise we don't have a return. Technically, I guess that would make this `@return bool|void`?

inc/blocks/namespace.php
```diff
@@ -691,3 +700,99 @@ function get_views( string $block_id, $args = [] ) {
...
+
+			if ( empty( $invalid ) ) {
+				return new Status( Status::COMPLETE, __( 'Experience Block conversion goals are valid', 'altis-analytics' ) );
+			}
+
+			// Check if this will block publishing.
+			// In some situations the validation check could be wrong so we shouldn't prevent publication.
+			$status = PublicationChecklist\should_block_publish() ? Status::INFO : Status::INCOMPLETE;
+
+			return new Status( $status, __( 'Experience Blocks with invalid conversion goals', 'altis-analytics' ), $invalid );
```
jazzsequence, line 795: This error status is phrased weird.

```suggestion
			return new Status( $status, __( 'Experience Blocks with invalid conversion goals found', 'altis-analytics' ), $invalid );
```

inc/blocks/personalization/components/variant-validation.js
```diff
@@ -0,0 +1,121 @@
...
+			} else {
+				noticeOperations.createErrorNotice(
+					( registeredGoal.args && registeredGoal.args.validation_message ) ||
+					sprintf( __( 'This variant does not meet the requirements for the "%s" conversion goal yet.', 'altis-analytics' ), registeredGoal.label )
+				);
+				updateBlockAttributes( clientId, { isValid: false } );
+			}
+		};
+
+		// Do an initial check incase the SSR component exists with no changes.
```
jazzsequence, line 86: Typo 
```suggestion
		// Do an initial check in case the SSR component exists with no changes.
```

jazzsequence (approved):
A couple minor, non blocking comments. Will do fairly extensive testing when I do QA, so not too worried about this at the moment.

inc/blocks/namespace.php
roborourke: That’s the return value of the closure in the array_filter function rather than this function

inc/blocks/namespace.php
roborourke: It’s optional but doesn’t matter too much. I think we’re started to do it more and making things explicit is generally good

PR #182, opened 2021-03-25
Fix XB post type capabilities

roborourke:
Setting the capabilities array when registering the post type meant those were the _only_ caps the post type object had, this caused fatals when trying to reference them elsewhere.


jazzsequence (approved):
👍

PR #187, opened 2021-03-26
Update text domains

roborourke:
After the merger of the 2 repos the text domains were left out, this fixes that!


jazzsequence (approved):
🎉🎉🎉

PR #223, opened 2021-06-18
Ensure status checks have text

roborourke:
I mistakenly thought an empty status check response would not show in the UI, we might need a means making that possible in the publication checklist plugin.


jazzsequence (changes requested):
Can we use a different word than "for" so we aren't ending with a preposition? Possibly `No Experience Blocks found to validate fallback content against`?

PR #233, opened 2021-07-29
Add full list of countries, and update select input display and logic

shadyvb:
This PR:

- Update the select input logic to accept a list from register_field, not just from available buckets.
- Allows restricting the input to predefined list only.
- Allows passing labels with the new list option above.
- Sorts options by bucket size if exists.

Note that the options are cached, so some cache busting will be needed to view the changes.


src/audiences/ui/components/rule.js
```diff
@@ -64,14 +64,24 @@ const RuleInput = props => {
 	} = props;
 
 	// Get the currently available values for the field.
-	const data = ( currentField.data && currentField.data.map( datum => datum.value ) ) || [];
+	const options = ( currentField.data && currentField.data.map( option => ( {
+		value: option.value,
+		label: ( option?.label ? `${ option.label } (${ option.value })` : option.value ) + String( option?.percent ? ` ~${ option.percent }%` : '' ),
```
roborourke, line 69: `option?.` is checking that `option` isn't `null` so it seems redundant for these checks. It'd make sense if we did the same for `option.value` above but at that point you could run a filter before the map on the `data` array to remove any undefined or falsey entries.

src/audiences/ui/components/rule.js
```diff
@@ -163,15 +173,29 @@ const RuleInput = props => {
 									} }
 								>
 									<option value="">{ __( 'Empty', 'altis-analytics' ) }</option>
-									{ data.map( datum => datum !== '' && (
+									{ topHits.length && (
```
roborourke, line 176: This should be an expression that results in a boolean, if `topHits.length` is 0 then "0" will be rendered:
```suggestion
									{ topHits.length > 0 && (
```

src/audiences/ui/components/rule.js
```diff
@@ -163,15 +173,29 @@ const RuleInput = props => {
...
+										<optgroup label={ __( 'Top hits', 'siemens' ) }>
```
roborourke, line 177: Wrong text domain 😛

src/audiences/ui/components/rule.js
```diff
@@ -64,14 +64,24 @@ const RuleInput = props => {
...
+	// Check if the filter doesn't allow free text.
+	const allowFreeText = ! currentField?.options?.disable_free_text;
 
 	// Get input refs to manage focus.
 	const inputEl = useRef( null );
 	const dropdownRef = useRef( null );
 
 	// Default to display of the dropdown if the value is in existing data or empty.
-	const defaultDropdownState = value === '' || data.indexOf( value ) >= 0;
+	const defaultDropdownState = value === '' || options.find( option => option.value === value );
```
roborourke, line 84: Hm, what's the reason for using `find` instead? This is intended to be a boolean value but `find` will return the matching item from the array I thought.

roborourke (changes requested):
Couple of minor things to change but this looks good. Can you update the docs in the README under the "Mapping Event Data" heading too? We'll need to add that to the main Analytics module docs too once we roll this update out as part of a 4.1.0 release.

src/audiences/ui/components/rule.js
shadyvb: I knew I should do that, just failed to remember why! :D will update.

src/audiences/ui/components/rule.js
shadyvb: 🤦

shadyvb:
```suggestion
										<optgroup label={ __( 'Top hits', 'altis-analytics' ) }>
```

src/audiences/ui/components/rule.js
shadyvb: True, needs casting here. It still worked as the casting happens below, and null is falsey.
```suggestion
	const defaultDropdownState = value === '' || !! options.find( option => option.value === value );
```

src/audiences/ui/components/rule.js
shadyvb: Hmmm, I think `option?.label` checks if `option` has a `label` value, rather than it being null. As not all values have a `label` but will always have a `value`.

shadyvb: .. or have I been using it wrong all this time ? :D

shadyvb: Yeah, seems like I was wrong, this is only useful in chained calls, `option.label` will return undefined anyway if it doesn't exist.

ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

shadyvb:
```suggestion
		label: ( option.label ? `${ option.label } (${ option.value })` : option.value ) + String( option.percent ? ` ~${ option.percent }%` : '' ),
```

shadyvb:
Updated the docs, back to you @roborourke

roborourke (approved):

PR #234, opened 2021-08-02
Add transform rules to personalization block

shadyvb:
Adds transform rules to personalization block to enable transforming blocks into P blocks.

- Allows multiple block transformation.
- Excludes P blocks and those that specify allowed list of parent blocks.


roborourke (approved):
Nailed it!

PR #251, opened 2021-08-19
Sync cached estimates and total and round to nearest hour

shadyvb:
Fixes sync issues between cached total audience count and fresh audience estimates. Synces both to the top of the hour, and minimize cached total TTL to one minute.

humanmade/altis-analytics#90

### DIFF
--- a/inc/audiences/namespace.php
+++ b/inc/audiences/namespace.php
@@ -8,6 +8,7 @@
 namespace Altis\Analytics\Audiences;
 
 use Altis\Analytics\Utils;
+use function Altis\Analytics\Utils\time_ago_in_milliseconds;
 use WP_Query;
 
 const COMPARISON_OPERATORS = [
@@ -430,6 +431,8 @@ function admin_enqueue_scripts() {
  * @return array|null
  */
 function get_estimate( array $audience ) : ?array {
+	$since = time_ago_in_milliseconds( '-1 week', HOUR_IN_SECONDS );
+
 	$query = [
 		'query' => [
 			'bool' => [
@@ -445,7 +448,7 @@ function get_estimate( array $audience ) : ?array {
 					[
 						'range' => [
 							'event_timestamp' => [
-								'gte' => Utils\milliseconds() - ( WEEK_IN_SECONDS * 1000 ),
+								'gte' => $since,
 							],
 						],
 					],
@@ -470,7 +473,7 @@ function get_estimate( array $audience ) : ?array {
 					'field' => 'event_timestamp',
 					'interval' => 6 * HOUR_IN_SECONDS * 1000, // 6 hour chunks.
 					'extended_bounds' => [
-						'min' => Utils\milliseconds() - ( WEEK_IN_SECONDS * 1000 ),
+						'min' => $since,
 						'max' => Utils\milliseconds(),
 					],
 				],
@@ -491,7 +494,7 @@ function get_estimate( array $audience ) : ?array {
 		return $cache;
 	}
 
-	$unique_count = get_unique_endpoint_count();
+	$unique_count = get_unique_endpoint_count( $since );
 	$result = Utils\query( $query );
 
 	if ( ! $result ) {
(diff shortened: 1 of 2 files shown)


inc/audiences/namespace.php
```diff
@@ -8,6 +8,7 @@
 namespace Altis\Analytics\Audiences;
 
 use Altis\Analytics\Utils;
+use function Altis\Analytics\Utils\time_ago_in_milliseconds;
```
roborourke, line 11: It's not enforced in our coding standards yet but we're discouraging the use of `use function` in favour of using the namespace and referencing the function as `Namespace\function()` so it's clearer where the function comes from in situ.

inc/audiences/namespace.php
```diff
@@ -430,6 +431,8 @@ function admin_enqueue_scripts() {
  * @return array|null
  */
 function get_estimate( array $audience ) : ?array {
+	$since = time_ago_in_milliseconds( '-1 week', HOUR_IN_SECONDS );
```
roborourke, line 434:
```suggestion
	$since = Utils\time_ago_in_milliseconds( '-1 week', HOUR_IN_SECONDS );
```

inc/audiences/namespace.php
```diff
@@ -533,10 +531,15 @@ function get_estimate( array $audience ) : ?array {
...
  * Get total unique endpoints for the past 7 days.
  *
+ * @param int $since Minimum date stamp in milliseconds.
  * @param bool $force_update Set to true to ignore the cache.
  * @return integer|null
  */
-function get_unique_endpoint_count( bool $force_update = false ) : ?int {
+function get_unique_endpoint_count( int $since = null, bool $force_update = false ) : ?int {
+	if ( $since === null ) {
+		$since = time_ago_in_milliseconds( '-1 week', HOUR_IN_SECONDS );
```
roborourke, line 540:
```suggestion
		$since = Utils\time_ago_in_milliseconds( '-1 week', HOUR_IN_SECONDS );
```

inc/utils/namespace.php
```diff
@@ -235,6 +235,31 @@ function milliseconds() : int {
...
+ * @param string $point_in_time strtotime-safe string, eg: '-1 week'
+ * @param integer $round_to Round the result to the nearest time block in seconds, eg: HOUR_IN_SECONDS.
+ *
+ * @return integer|null
+ */
+function time_ago_in_milliseconds( string $point_in_time, int $round_to = 0 ) : ?int {
+	$time_ago = strtotime( $point_in_time );
+
+	if ( ! $time_ago ) {
+		trigger_error( 'Analytics: Point in time string cannot be resolved.', E_USER_WARNING );
```
roborourke, line 250: We could make the error message a bit friendlier by adding in the value that was passed, would likely help with debugging.

inc/utils/namespace.php
```diff
@@ -235,6 +235,31 @@ function milliseconds() : int {
...
 
+/**
+ * Get past point in time in milliseconds, optionally rounded to the nearest time block.
+ *
+ * @param string $point_in_time strtotime-safe string, eg: '-1 week'
+ * @param integer $round_to Round the result to the nearest time block in seconds, eg: HOUR_IN_SECONDS.
+ *
+ * @return integer|null
+ */
+function time_ago_in_milliseconds( string $point_in_time, int $round_to = 0 ) : ?int {
```
roborourke, line 246: Although time ago is our use for this right now it's really a milliseconds wrapper for `strtotime()` which is certainly useful here, I could technically pass in any future date or "+1 week" for example. Maybe the function can be named `str_to_milliseconds` or something along those lines to indicate it's function.

Alternatively if we want to ensure the point in time is in the past then the function should check for that.

roborourke (changes requested):
Couple of minor updates but good catch on the reason for the error. Also makes sense to keep cache times standardised on 1 minute too. Nice work.

inc/audiences/namespace.php
shadyvb:
```suggestion
```

inc/utils/namespace.php
shadyvb: Yeah, good catch, generalizing it makes more sense.

roborourke (approved):

roborourke (approved):

PR #259, opened 2021-08-19
Use polyfills.io to fix IE compat issue with ES6 features

shadyvb:
Polyfills IE with ES6 features, including Promises which is required by Analytics script. Opted for the full feature set to future-proof it.

Notes:
- ~I was unable to see the audience changes within audience manager, however I verified the entries were added via Elastic.~ They actually show up, it was just ~0% of the total still 🤦 
- There is a syntax error with the consent API JS file, which I'm pushing a PR for shortly. > https://github.com/humanmade/consent-api-js/pull/14

humanmade/altis-analytics#136

### DIFF
--- a/inc/namespace.php
+++ b/inc/namespace.php
@@ -247,6 +247,15 @@ function enqueue_scripts() {
 	 */
 	$consent_cookie_prefix = apply_filters( 'wp_consent_cookie_prefix', 'wp_consent' );
 
+	// Use polyfills.io to fix IE compat issues, only polyfilling features where not supported.
+	wp_enqueue_script(
+		'polyfill.io',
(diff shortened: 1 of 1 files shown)


roborourke:
Any idea what the performance overhead for requesting this is? For browsers that don't need it that is. Also should it be a dependency of the analytics.js script?

I came across this pattern for avoiding requesting it at all on modern browsers (with the exception of Safari 10.1) https://3perf.com/blog/polyfills/#modulenomodule

If we add the `nomodule` attribute to the script tag it'll only load in old browsers. I'm not aware currently of any other browsers where the current bundle is failing. That would take a bunch more research.

roborourke (approved):

PR #276, opened 2021-09-20
Search specific indexes rather than wildcard query

roborourke:
This improves the performance on large datasets by only querying relevant indexes but also adding request caching where appropriate to improve look ups on indexes with static data e.g. those older than one day.

### DIFF
--- a/inc/utils/namespace.php
+++ b/inc/utils/namespace.php
@@ -238,6 +238,17 @@ function query( array $query, array $params = [], string $path = '_search', stri
 		$path
 	) );
 
+	// Elasticsearch supports up to 4kb URLs by default so we can revert to the wildcard in this instance.
+	// Note it should never happen as 90 indexes would be 1889 bytes total.
+	if ( strlen( $url ) > 4 * 1024 ) {
+		$url = add_query_arg( $params, sprintf(
+			'%s/%s/%s',
+			get_elasticsearch_url(),
+			'analytics-*',
+			$path
+		) );
+	}
+
 	// Escape the URL to ensure nothing strange was passed in via $path.
 	$url = esc_url_raw( $url );
 
--- a/inc/utils/namespace.php
+++ b/inc/utils/namespace.php
@@ -166,7 +166,7 @@ function get_indices() : array {
 
 	if ( wp_remote_retrieve_response_code( $indices_response ) !== 200 ) {
 		trigger_error( sprintf(
-			"Analytics: ElasticSearch index deletion failed:\n%s",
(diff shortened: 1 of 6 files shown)


roborourke:
cc @kevinlangleyjr - after cache kicked in I found between a 2,000% - 10,000% performance increase for some queries like audiences with this approach

inc/utils/namespace.php
```diff
@@ -155,8 +194,49 @@ function query( array $query, array $params = [], string $path = '_search', stri
...
+			break;
+		}
+	}
+
 	// Get URL.
-	$url = add_query_arg( $params, get_elasticsearch_url() . '/analytics*/' . $path );
+	$url = add_query_arg( $params, sprintf(
+		'%s/%s/%s',
+		get_elasticsearch_url(),
+		implode( ',', $index_paths ),
```
shadyvb, line 237: Should we add any guards against too many selected indices here ?

shadyvb (changes requested):

inc/utils/namespace.php
```diff
@@ -142,6 +142,45 @@ function get_elasticsearch_version() : ?string {
...
+			'Analytics: Could not fetch analytics indexes: %s',
+			// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+			$indices_response->get_error_message()
+		), E_USER_WARNING );
+		return [];
+	}
+
+	if ( wp_remote_retrieve_response_code( $indices_response ) !== 200 ) {
+		trigger_error( sprintf(
+			"Analytics: ElasticSearch index deletion failed:\n%s",
```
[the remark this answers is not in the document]
roborourke: 🤦‍♂️  good catch

inc/utils/namespace.php
roborourke: Yeah maybe. Elasticsearch allows up to 4kb in the URL - with 90 analytics indexes it's 1800 ascii characters (20 characters per index name) plus 89 commas.

roborourke: I'm not sure it's worth it looking at it, we cap the number of indexes at 90 and you'd get an error reported and a `null` value returned.

shadyvb: If it exceeds the cap we can query for the wildcard instead, otherwise we'd still have the error/null return.

roborourke: True, I've updated it for that. I suppose if someone really really wanted to they could unschedule the index cleanup routine.

shadyvb (approved):

roborourke:
Thanks @shadyvb

PR #281, opened 2021-09-22
Exclude bot traffic from analytics data

### LINKED ISSUE #23: Ensure search engines are excluded
Ideally we shouldn't include search engines. We need to do this client side in `analytics.js` by checking `window.navigator.userAgent` if present.

This article lists all the most popular search engines and their User Agent strings along with a regex match you can use for them https://perishablepress.com/list-all-user-agents-top-search-engines/

**Acceptance Criteria:**
- [ ] Research: If search engine bot traffic is being counted
- [ ] Search engine traffic should not be included in the audience estimation and XB view analytics
- [ ] Discuss whether we want the user to know the split between the bot traffic and actual user traffic: validate whether this piece of data is of use to the client

roborourke:
Adds the ability to enable bots too with an attribute for filtering bot traffic.

Fixes #23

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ The plugin provides a few hooks for you to control the default endpoint data and
 
 **`altis.analytics.consent_enabled <bool>`**
 
-This defaults to if the WP Consent API plugin or a derivative is installed and active by checking if the constant `WP_CONSENT_API_URL` is defined.
+This defaults to true if the WP Consent API plugin or a derivative is installed and active by checking if the constant `WP_CONSENT_API_URL` is defined.
 
 **`altis.analytics.data.endpoint <array>`**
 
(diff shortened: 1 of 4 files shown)


shadyvb (commented):
@roborourke do we want to use an external maintained library like https://github.com/DamonOehlman/detect-browser/releases ( with no dependencies ) to manage the detection bit, so we offload the maintenance efforts instead ?

roborourke:
@shadyvb that's a good point. but I don't really want to add more dependencies, in fact I'd prefer to have none in this particular script to make it as lean as possible.

Saying that we're already using the `ua-parser-js` library, though the one you linked looks to have a smaller package size, if we can get the same data out of it we could switch to that instead. I'll look into it.

roborourke:
Actually `detect-browser` doesn't provide enough information, Pinpoint wants more info than just browser, version and OS.

So, sticking with this implementation!

shadyvb (approved):

PR #284, opened 2021-09-30
Fix XB compatibility with WP 5.8

### LINKED ISSUE #283: XBs broken in WP 5.8
Experience blocks are broken in WP 5.8 due to requiring registration in JS as well as server side. The `altis/shim` block is causing errors as a result, it is technically only required server side but it can be added and hidden.

roborourke:
WP 5.8 requires that blocks are registered in JS now as well, the shim block for server side validation was broken as a result.

Fixes #283


rmccue (approved):
sure why not

PR #292, opened 2021-11-12
Add AB Test block web component

roborourke:
Fixes humanmade/altis-analytics#193

### DIFF
--- a/src/experiments.js
+++ b/src/experiments.js
@@ -23,7 +23,7 @@ function isVisible( element ) {
  */
 class Test extends HTMLElement {
 
-	storageKey = '_altis_tests';
+	storageKey = '_altis_ab_tests';
 
 	get testId() {
 		return this.getAttribute( 'test-id' );
@@ -38,11 +38,15 @@ class Test extends HTMLElement {
 	}
 
 	get trafficPercentage() {
-		return this.getAttribute( 'traffic-percentage' );
+		return parseFloat( this.getAttribute( 'traffic-percentage' ) );
 	}
 
 	get variants() {
 		return JSON.parse( this.getAttribute( 'variants' ) ) || [];
+	}
+
+	get variantWeights() {
+		return ( JSON.parse( this.getAttribute( 'variant-weights' ) ) || [] ).map( parseFloat );
 	}
 
 	get fallback() {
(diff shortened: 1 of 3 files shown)


src/experiments.js
```diff
@@ -88,15 +92,64 @@ class Test extends HTMLElement {
...
+		// that it's a valid index.
+		if (
+			typeof currentTests[ testId ] !== 'undefined' &&
+			currentTests[ testId ] !== false &&
+			currentTests[ testId ] < this.variants.length
+		) {
+			variantId = currentTests[ testId ];
+		} else if ( currentTests[ testId ] === false ) {
+			return variantId;
+		} else {
```
shadyvb, line 112: Curious to see if we can shorten this is a bit, just to see if there is a case I couldn't account for:
```suggestion
		let variantId = currentTests[ testId ] || false;
		if ( ! variantId ) {
```

since :
```js
const values = ['x'];
// All the below are true
console.log( values[undefined] === undefined );
console.log( values['string'] === undefined );
console.log( values[false] === undefined );
console.log( values[10] === undefined );
```

shadyvb (approved):
LGTM, just a comment about shortening a condition.

src/experiments.js
roborourke: I think this does do the job but it's losing 2 things:

- basic `currentTests[ testId ]` value validation from `currentTests[ testId ] < this.variants.length`
- early return if `currentTests[ testId ]` is set but the value is `false`

We need to differentiate between undefined and `false` in terms of the value - `false` means the visitor is excluded from the test, while `undefined` means determine whether they should be in the test, then if so which variant they should get.

Maybe something like:

```suggestion
		let variantId = false;
		// Test variant can be 0 so check for not undefined and not strictly false and
		// that it's a valid index.
		let variantId = currentTests[ testId ];
		if ( variantId === false ) {
			return variantId;
		} else if ( typeof variantId === 'undefined' || variantId >= this.variants.length ) {
```

roborourke: I'll leave it this way for now but I'll add an issue for a refactor of this file. It could do with splitting out and consolidating further really.

roborourke:
Issue for further refactoring: https://github.com/humanmade/aws-analytics/issues/293

src/experiments.js
shadyvb: Sure, it's not a blocker. And I didn't factor the length case. 👍

PR #294, opened 2021-11-30
Fix getting variant traffic percentage for post as an array with empty string

robindevitt:
When updating a post or page the variant traffic percentage would cause an error. 

The function `get_ab_test_variant_traffic_percentage_for_post` would return an an empty string as defined as an array.

The post meta is now checked first if it's an array. It's returned as an array if it is, else an empty array is returned

### DIFF
--- a/inc/experiments/namespace.php
+++ b/inc/experiments/namespace.php
@@ -542,10 +542,11 @@ function get_ab_test_traffic_percentage_for_post( string $test_id, int $post_id 
  *
  * @param string $test_id The test ID.
  * @param int $post_id Post ID to get test data for.
- * @return float[] Array of percentages.
+ * @return array $traffic_percent array
  */
 function get_ab_test_variant_traffic_percentage_for_post( string $test_id, int $post_id ) : array {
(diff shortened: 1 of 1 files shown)


inc/experiments/namespace.php
```diff
@@ -542,10 +542,11 @@ function get_ab_test_traffic_percentage_for_post( string $test_id, int $post_id
  *
  * @param string $test_id The test ID.
  * @param int $post_id Post ID to get test data for.
- * @return float[] Array of percentages.
+ * @return array $traffic_percent array
```
roborourke, line 545: This bit was fine fwiw - because it's in a doc block it won't affect how the code runs but can give editors a bit more of an idea of what the function should return. I think this kind of thing is possible in PHP8, so you could do type declarations like `float[]` meaning an array of floating point numbers.

roborourke (approved):
Nice one :)

inc/experiments/namespace.php
robindevitt: Thank you Rob and noted, Should I restore the doc block changes?

PR #295, opened 2021-12-02
Add ability to preview different A/B test block variants

robindevitt:
Giving users the ability to preview A/B test block variants as they add a block.

Issue https://github.com/humanmade/altis-analytics/issues/194

### DIFF
--- a/inc/blocks/ab-test/components/variant-panel.js
+++ b/inc/blocks/ab-test/components/variant-panel.js
@@ -84,13 +84,15 @@ const VariantPanel = ( { index, postId, title, variant, variants, onMouseDown } 
 					} );
 				} }
 			/>
-			<Preview
-				href={ `/?p=${ postId }&set_test=ab_block_${ postId }:${ index }` }
-				target="_ab_test_preview"
-			>
-				{ __( 'Preview', 'altis-analytics' ) }
-				<Icon icon="external" />
-			</Preview>
+			{ !! postId && (
+				<Preview
+					href={ `/?p=${ postId }&set_test=ab_block_${ postId }:${ index }` }
+					target="_ab_test_preview"
+				>
+					{ __( 'Preview', 'altis-analytics' ) }
+					<Icon icon="external" />
+				</Preview>
+			) }
 		</PanelBody>
 	);
 };
--- a/inc/blocks/ab-test/components/variant-panel.js
+++ b/inc/blocks/ab-test/components/variant-panel.js
@@ -42,7 +42,7 @@ const Preview = styled( Button ).attrs( {
  * @param {React.ReactNode} props.placeholder Placeholder text for the variant title while loading.
  * @returns {React.ReactNode} Variant settings panel.
  */
-const VariantPanel = ( { index, postId, title, variant, variants, onMouseDown } ) => {
+const VariantPanel = ( { index, blockId, postId, title, variant, variants, onMouseDown } ) => {
 	const { updateBlockAttributes } = useDispatch( 'core/block-editor' );
 
 	const defaultPercentage = 100 / variants.length;
@@ -86,7 +86,7 @@ const VariantPanel = ( { index, postId, title, variant, variants, onMouseDown } 
 			/>
 			{ !! postId && (
 				<Preview
-					href={ `/?p=${ postId }&set_test=ab_block_${ postId }:${ index }` }
+					href={ `/?p=${ postId }&set_test=${ blockId }&i=${ index }` }
 					target="_ab_test_preview"
 				>
 					{ __( 'Preview', 'altis-analytics' ) }
--- a/inc/blocks/ab-test/components/variant-panel.js
+++ b/inc/blocks/ab-test/components/variant-panel.js
@@ -52,7 +52,7 @@ const VariantPanel = ( { index, blockId, postId, title, variant, variants, onMou
 	const minPercentage = variants.length > 1 ? 0 : 100;
 
 	// Get A/B Test Block Id
-	const ab_post = useSelect( select => {
+	const abPost = useSelect( select => {
 		return select( 'analytics/xbs' ).getPost( blockId );
 	}, [ blockId ] );
 
@@ -92,9 +92,9 @@ const VariantPanel = ( { index, blockId, postId, title, variant, variants, onMou
 					} );
 				} }
 			/>
-			{ !! ab_post && (
+			{ !! abPost && (
 				<Preview
-					href={ `/?p=${ postId }&set_test=test_xb_${ ab_post.id }:${ index }` }
+					href={ `/?p=${ postId }&set_test=test_xb_${ abPost.id }:${ index }` }
 					target="_ab_test_preview"
 				>
 					{ __( 'Preview', 'altis-analytics' ) }
--- a/inc/blocks/ab-test/components/variant-panel.js
+++ b/inc/blocks/ab-test/components/variant-panel.js
@@ -94,7 +94,7 @@ const VariantPanel = ( { index, blockId, postId, title, variant, variants, onMou
 			/>
 			{ !! ab_post && (
 				<Preview
-					href={ `/?p=${ postId }&set_test=ab_block_${ ab_post.id }:${ index }` }
+					href={ `/?p=${ postId }&set_test=test_xb_${ ab_post.id }:${ index }` }
 					target="_ab_test_preview"
 				>
 					{ __( 'Preview', 'altis-analytics' ) }
--- a/inc/blocks/ab-test/components/variant-panel.js
+++ b/inc/blocks/ab-test/components/variant-panel.js
@@ -42,7 +42,7 @@ const Preview = styled( Button ).attrs( {
  * @param {React.ReactNode} props.placeholder Placeholder text for the variant title while loading.
  * @returns {React.ReactNode} Variant settings panel.
  */
-const VariantPanel = ( { postId, title, variant, variants, onMouseDown } ) => {
+const VariantPanel = ( { index, postId, title, variant, variants, onMouseDown } ) => {
 	const { updateBlockAttributes } = useDispatch( 'core/block-editor' );
 
 	const defaultPercentage = 100 / variants.length;
@@ -85,7 +85,7 @@ const VariantPanel = ( { postId, title, variant, variants, onMouseDown } ) => {
 				} }
 			/>
 			<Preview
-				href={ `/?p=${ postId }&set_test=ab_block_${ postId }:${ variant.clientId }` }
+				href={ `/?p=${ postId }&set_test=ab_block_${ postId }:${ index }` }
 				target="_ab_test_preview"
 			>
 				{ __( 'Preview', 'altis-analytics' ) }
--- a/inc/blocks/ab-test/components/variant-panel.js
+++ b/inc/blocks/ab-test/components/variant-panel.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import styled from 'styled-components';
 
 import GoalPicker from '../../personalization/components/goal-picker';
 
@@ -6,9 +7,32 @@ const {
 	PanelBody,
 	TextControl,
 	RangeControl,
+	Button,
+	Icon,
 } = wp.components;
 const { useDispatch } = wp.data;
 const { __ } = wp.i18n;
+
+const Preview = styled( Button ).attrs( {
+	isLink: true,
+} )`
+	text-align: right;
+	flex: 0;
+	align-self: flex-end;
+	font-size: inherit;
+
+	.dashicon {
+		width: auto;
+		font-size: inherit;
+		line-height: inherit;
+		vertical-align: baseline;
+	}
+
+	svg.dashicon, .dashicon svg {
+		width: 0.9rem;
+		margin-left: 2px;
+	}
+`;
 
 /**
  * Variant settings panel.
@@ -18,7 +42,7 @@ const { __ } = wp.i18n;
  * @param {React.ReactNode} props.placeholder Placeholder text for the variant title while loading.
  * @returns {React.ReactNode} Variant settings panel.
  */
-const VariantPanel = ( { title, variant, variants, onMouseDown } ) => {
+const VariantPanel = ( { postId, title, variant, variants, onMouseDown } ) => {
 	const { updateBlockAttributes } = useDispatch( 'core/block-editor' );
 
 	const defaultPercentage = 100 / variants.length;
@@ -60,6 +84,13 @@ const VariantPanel = ( { title, variant, variants, onMouseDown } ) => {
 					} );
 				} }
 			/>
+			<Preview
+				href={ `/?p=${ postId }&set_test=ab_block_${ postId }:${ variant.clientId }` }
+				target="_ab_test_preview"
+			>
+				{ __( 'Preview', 'altis-analytics' ) }
+				<Icon icon="external" />
+			</Preview>
 		</PanelBody>
 	);
 };
--- a/inc/blocks/ab-test/components/variant-panel.js
+++ b/inc/blocks/ab-test/components/variant-panel.js
@@ -10,7 +10,10 @@ const {
 	Button,
 	Icon,
 } = wp.components;
-const { useDispatch } = wp.data;
+const {
+	useDispatch,
+	useSelect,
+} = wp.data;
 const { __ } = wp.i18n;
 
 const Preview = styled( Button ).attrs( {
@@ -48,6 +51,11 @@ const VariantPanel = ( { index, blockId, postId, title, variant, variants, onMou
 	const defaultPercentage = 100 / variants.length;
 	const minPercentage = variants.length > 1 ? 0 : 100;
 
+	// Get A/B Test Block Id
+	const ab_post = useSelect( select => {
+		return select( 'analytics/xbs' ).getPost( blockId );
+	}, [ blockId ] );
+
 	return (
 		<PanelBody title={ title } onMouseDown={ onMouseDown }>
 			<TextControl
@@ -84,9 +92,9 @@ const VariantPanel = ( { index, blockId, postId, title, variant, variants, onMou
 					} );
 				} }
 			/>
-			{ !! postId && (
+			{ !! ab_post && (
 				<Preview
-					href={ `/?p=${ postId }&set_test=${ blockId }&i=${ index }` }
+					href={ `/?p=${ postId }&set_test=ab_block_${ ab_post.id }:${ index }` }
 					target="_ab_test_preview"
 				>
 					{ __( 'Preview', 'altis-analytics' ) }
--- a/inc/blocks/ab-test/edit.css
+++ b/inc/blocks/ab-test/edit.css
@@ -1,13 +1,28 @@
+.ab-test-xb-preview__tabs {
+    margin: 0.5rem 0rem;
+    line-height: 1em;
+}
+
 .ab-test-xb-preview__tab {
-	background-color: #4667de;
+	background: #4667de;
 	padding: 0.25rem 0.5rem;
 	font-size: 0.75rem;
-	margin-right: 0.25rem;
+	margin: 0 0.25rem 0 0;
+	font-family:
+		-apple-system,
+		BlinkMacSystemFont,
+		"Segoe UI",
+		Roboto,
+		Oxygen-Sans,
+		Ubuntu,
+		Cantarell,
+		"Helvetica Neue",
+		sans-serif;
 }
 
 .ab-test-xb-preview__tab.active,
 .ab-test-xb-preview__tab:hover,
 .ab-test-xb-preview__tab:focus {
-	background-color: #152a4e;
+	background: #152a4e;
 	cursor: pointer;
 }
--- a/inc/blocks/ab-test/edit.css
+++ b/inc/blocks/ab-test/edit.css
@@ -1,7 +1,7 @@
 button.ab-test-xb-preview__tab {
 	background-color: #4667de;
 	padding: 0.25rem 0.5rem;
-    font-size: 0.75rem;
+	font-size: 0.75rem;
 	margin-right: 0.25rem;
 }
 
--- a/inc/blocks/ab-test/edit.css
+++ b/inc/blocks/ab-test/edit.css
@@ -1,5 +1,5 @@
 .ab-test-xb-preview__tabs {
-	margin: 0.5rem 0 !important;
+	margin: 0.5rem 0;
 	line-height: 1;
 }
 
--- a/inc/blocks/ab-test/edit.css
+++ b/inc/blocks/ab-test/edit.css
@@ -1,13 +1,13 @@
 .ab-test-xb-preview__tabs {
-    margin: 0.5rem 0rem;
-    line-height: 1em;
+	margin: 0.5rem 0 !important;
+	line-height: 1;
 }
 
 .ab-test-xb-preview__tab {
 	background: #4667de;
 	padding: 0.25rem 0.5rem;
 	font-size: 0.75rem;
-	margin: 0 0.25rem 0 0;
+	margin: 0 0.25rem 0.25rem 0;
 	font-family:
 		-apple-system,
 		BlinkMacSystemFont,
--- a/inc/blocks/ab-test/edit.css
+++ b/inc/blocks/ab-test/edit.css
@@ -1,13 +1,13 @@
-button.ab-test-xb-preview__tab {
+.ab-test-xb-preview__tab {
 	background-color: #4667de;
 	padding: 0.25rem 0.5rem;
 	font-size: 0.75rem;
 	margin-right: 0.25rem;
 }
 
-button.ab-test-xb-preview__tab.active,
-button.ab-test-xb-preview__tab:hover,
-button.ab-test-xb-preview__tab:focus {
+.ab-test-xb-preview__tab.active,
+.ab-test-xb-preview__tab:hover,
+.ab-test-xb-preview__tab:focus {
 	background-color: #152a4e;
 	cursor: pointer;
 }
--- a/inc/blocks/ab-test/edit.css
+++ b/inc/blocks/ab-test/edit.css
@@ -1,0 +1,13 @@
+button.ab-test-xb-preview__tab {
+	background-color: #4667de;
+	padding: 0.25rem 0.5rem;
+    font-size: 0.75rem;
+	margin-right: 0.25rem;
+}
+
+button.ab-test-xb-preview__tab.active,
+button.ab-test-xb-preview__tab:hover,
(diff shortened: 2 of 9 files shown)


src/experiments.js
```diff
@@ -616,3 +616,15 @@ window.Altis.Analytics.onReady( () => {
 // Fire a ready event once userland API has been exported.
 const readyEvent = new CustomEvent( 'altis.experiments.ready' );
 window.dispatchEvent( readyEvent );
+
+document.querySelectorAll( '.ab-test-xb-preview' ).forEach( function ( xb ) {
+	let templates = document.querySelectorAll( 'template[data-parent-id="' + xb.dataset.clientId + '"]' );
```
robindevitt, line 621: My editor kept throwing an error for this and suggested the use of "let". Should it be a var instead?

inc/blocks/ab-test/register.php
```diff
@@ -113,6 +113,17 @@ function render_block( array $attributes, ?string $inner_content = '' ) : string
...
+	if ( is_preview() ) {
+		return sprintf(
+			'%s<div class="ab-test-xb-preview %s" data-client-id="%s"><div class="ab-test-xb-preview__tabs" data-client-tabs="%s"></div><div class="ab-test-xb-preview__content" data-client-content="%s"></div></div>', 
+			$inner_content,
+			$class_name, 
+			$client_id,
+			$client_id,
+			$client_id
+		);
+	}
```
robindevitt, line 125: I have started this and I wanted to have the tabs and the content identifiable as to which tabs are with which content. Is there a better way to go around this that I may not have to repeat the `$client_id` multiple times?

roborourke: The value for the client ID is already available on the outer div, I'm not sure what the benefit in repeating it would be?

To answer your question though take a look at the sprintf docs: https://www.php.net/manual/en/function.sprintf.php

You can reuse arguments by using the `%<arg position>$s` syntax, so for the first argument you pass `%s` would become `%1$s`, then the second would be `%2$s`. When used that way you can reference them by number as often as you want.

src/experiments.js
roborourke: If you're ever in doubt check caniuse.com, eg. https://caniuse.com/let

In this case `let` is fine as we don't support IE older than 10, and as this is for admin users only we don't even support IE11 there.

I would split this out into a separate JS file though that's only loaded when we're previewing content to keep the end user JS as small and lean as it can be.

src/experiments.js
```diff
@@ -616,3 +616,15 @@ window.Altis.Analytics.onReady( () => {
...
+	let tabCount = 1;
+	for ( let i = 0; i < templates.length; i++ ) {
+		document.querySelector( '.ab-test-xb-preview__tabs[data-client-tabs="' + xb.dataset.clientId + '"]' ).innerHTML += ( '<button data-tab="' + tabCount + '" class="variant">Variant ' + tabCount + '</button>' );
+		let innherHTML = templates[i].innerHTML;
+		document.querySelector( '.ab-test-xb-preview__content[data-client-content="' + xb.dataset.clientId + '"]' ).innerHTML += ( '<div class="variant-content">' + innherHTML + '</div>' );
```
roborourke, line 626: We already have the content we need in each `<template>` component, if you look further up in the file you can see that we swap the contents out using something like the following:

```js

// then in the on click handler:
tabContent.innerHTML = ''; // empty the content
const variant = templates[i].content.cloneNode( true );
tabContent.appendChild( variant );
```

Note that `querySelector()` and `querySelectorAll()` are also methods of any `Element` reference so they can be scoped to specific elements rather than always querying from `document`: https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelector

inc/blocks/ab-test/components/variant-panel.js
```diff
@@ -60,6 +84,13 @@ const VariantPanel = ( { title, variant, variants, onMouseDown } ) => {
 					} );
 				} }
 			/>
+			<Preview
+				href={ `/?p=${ postId }&set_test=ab_block_${ postId }:${ index }` }
+				target="_ab_test_preview"
+			>
+				{ __( 'Preview', 'altis-analytics' ) }
+				<Icon icon="external" />
+			</Preview>
```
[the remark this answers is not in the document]
roborourke: The `!!` ensure that the value of `postId` is not just falsy but definitely a boolean. In some cases say if it was a `0` and we just wrote `postId && ( ... )` then it would output "0" as a string.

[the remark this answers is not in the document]
robindevitt: This does work with a draft post.

I do agree with your suggestion on adding `preview=xyz` to the preview URL's. Especially for the need to have the url to set a default state of a tab. Example were if the user clicked preview from "Variant C" when they previewed, "Variant C" should be active.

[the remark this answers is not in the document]
roborourke: Cool, the set_test param can be used to set the default tab then

src/experiments.js
robindevitt: Oooh that's neat! I always thought https://caniuse.com/ was for CSS, Thank you, I now know that it can be used for front-end languages!

robindevitt: For the separate file is there a particular method to adding it and including it in the build process? I have tried to get it to work but feel I may be missing something furhter.

I have tried a `wp_enqueue_script` in `aws-analytics/inc/blocks/ab-test/register.php` no luck. I have also tried following the setup of the other files but no success for me yet.

inc/blocks/ab-test/register.php
robindevitt: Will rework this and remove the items if they not needed any longer

roborourke: To explain a bit further you'd have an initial loop to go through all AB test blocks on the page from where you can get the current block and the client ID, then you'd get the templates associated with it and loop over those, adding the tabs and click handlers.

src/experiments.js
roborourke: You can push work in progress and I can comment on that or we can jump on a huddle. It might not even strictly need to be built.

robindevitt:
- Files can be enqueued. For this task the file for the A/B test preview it's suggested the file be included in the `is_preview` check for the rendering of the content 

- Files for the build process can be included in `webpack.config.js` and the value that is used in the filenames in the build folder: eg: `analytics.6e300f1c.js` for caching purposes

- follow lint suggestions when completing WIP

src/ab-test-preview.js
```diff
@@ -0,0 +1,36 @@
+import { getLetter } from '../src/utils';
+
+document.querySelectorAll( '.ab-test-xb-preview' ).forEach( function ( xb ) {
+
+	let templates = document.querySelectorAll( 'template[data-parent-id="' + xb.dataset.clientId + '"]' );
```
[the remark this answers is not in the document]
robindevitt: Thank you

src/ab-test-preview.js
```diff
@@ -0,0 +1,36 @@
...
+			const variant = templates[i].content.cloneNode( true );
+			tab.className = 'ab-test-xb-preview__tab active';
+			tabContent.innerHTML = '';
+			tabContent.appendChild( variant );
+		} else {
+			tab.className = 'ab-test-xb-preview__tab';
+		}
+
+		tab.setAttribute( 'data-tab', i );
+		tab.innerHTML = 'Variant ' + getLetter( i );
```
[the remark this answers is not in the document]
robindevitt: Ill mark this resolved and comment on the changes I made for a suggestion on a final item

src/ab-test-preview.js
```diff
@@ -0,0 +1,36 @@
...
+
+	for ( let i = 0; i < templates.length; i++ ) {
+
+		const tab = document.createElement( 'button' );
+
+		if ( i === 0 ){
+			const variant = templates[i].content.cloneNode( true );
+			tab.className = 'ab-test-xb-preview__tab active';
+			tabContent.innerHTML = '';
+			tabContent.appendChild( variant );
```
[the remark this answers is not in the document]
robindevitt: I had made some changes to this section so Im not entirely sure if it will still be relevant. But I currently have the first tab being active. I have an issue when setting a preview for the second tab, then the first tab is also active.

inc/blocks/ab-test/components/variant-panel.js
```diff
@@ -60,6 +84,15 @@ const VariantPanel = ( { title, variant, variants, onMouseDown } ) => {
 					} );
 				} }
 			/>
+			{ !! postId && (
+				<Preview
+					href={ `/?p=${ postId }&set_test=${ blockId }&i=${ index }` }
```
roborourke, line 89: This link format isn't quite right, this is the regex currently used to check the value: `(utm_campaign|set_test)=test_${ this.testIdWithPost }:(\d+)`, where `testIdWithPost` would be `xb_<xb post id>`.

src/ab-test-preview.js
```diff
@@ -0,0 +1,44 @@
...
+document.querySelectorAll( '.ab-test-xb-preview' ).forEach( function ( xb ) {
+
+	const templates = document.querySelectorAll( 'template[data-parent-id="' + xb.dataset.clientId + '"]' );
+	const tabContainer = xb.querySelector( '.ab-test-xb-preview__tabs' );
+	const tabContent = xb.querySelector( '.ab-test-xb-preview__content' );
+
+	for ( let i = 0; i < templates.length; i++ ) {
+
+		const tab = document.createElement( 'button' );
+		if ( i === 0 || i === abIndex && xb.dataset.clientId === abtest ){
```
roborourke, line 17: I still think it's worth avoiding the duplication with the click handler code and handling this logic to set the default tab after looping over the templates. You have all the info you need to select the right tab and call `.click()` on it as it'll either be index 0 or the value of `abIndex` if the block client ID matches.

inc/blocks/ab-test/edit.js
```diff
@@ -212,7 +213,10 @@ const Edit = ( {
 				{ variants.map( ( variant, index ) => (
 					<VariantPanel
 						key={ `variant-settings-${ variant.clientId }` }
+						blockId= { attributes.clientId }
```
roborourke, line 216: I get where you were going with this now. It does avoid needing to make an API request to get the XB's post ID.

src/ab-test-preview.js
```diff
@@ -0,0 +1,44 @@
+import { getLetter } from '../src/utils';
+
+const queryString = window.location.search;
+const urlParams = new URLSearchParams(queryString);
+const abtest = urlParams.get( 'set_test' );
+const abIndex = parseFloat( urlParams.get( 'i' ) );
```
roborourke, line 6: What was the reason for choosing `parseFloat()` over `parseInt()` here?

roborourke (commented):
Few more notes, be sure to go through and fix the linter errors shown on the files tab for this PR too.

Keep a close eye on the code style / spacing too to keep it consistent with the existing code.

inc/blocks/ab-test/components/variant-panel.js
roborourke: I understand what you're doing here now, so it's achieving the same thing more or less but avoiding the need to fetch the XB post in the back end just to get the ID. I think it's ok to stick with this.

src/ab-test-preview.js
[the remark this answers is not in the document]
roborourke: Still relevant but will need to handle getting the right tab item to fire the click event on rather than just using the first one.

src/ab-test-preview.js
robindevitt: In my latest commit ( I just did a commit now ) I made some further changes to the click handler

src/ab-test-preview.js
robindevitt: In my latest commit I have changed this approach with the existing preview link structure. Only really fighting an issue to extract the relevant info from the URL. I'll comment on the section I have an issue with.

src/ab-test-preview.js
```diff
@@ -0,0 +1,58 @@
+import { getLetter } from '../src/utils';
+
+const queryString = window.location.search;
+const urlParams = new URLSearchParams(queryString).get( 'set_test' );
+const abtest = urlParams.get( 'set_test' ).replace( 'ab_block_', '' ).split( ':', 2 );
```
robindevitt, line 5: @roborourke I am trying to use the way that you had originally suggested. I am fighting an issue for this when the using is using a preview URL for the page and not the AB Test URL I then have an issue around the `abtest` as the preview URL doesn't have a value for the `replace`.

robindevitt:
Hey @roborourke 

I have done final changes and cleaned up the HM Linter issues. 

Any open "conversations" I will be sure to resolve them later with final comments.

I have submit the request for review when you have a moment. Thank you.

inc/blocks/ab-test/components/variant-panel.js
```diff
@@ -1,15 +1,42 @@
...
+	Icon,
 } = wp.components;
-const { useDispatch } = wp.data;
+const {
+	useDispatch,
+	useSelect,
+} = wp.data;
 const { __ } = wp.i18n;
 
+const Preview = styled( Button ).attrs( {
```
roborourke, line 19: I think if this is copied from the component in the titles AB test code we could make into a shared component that can be reused. Not essential but good practice whenever possible.

inc/blocks/ab-test/components/variant-panel.js
```diff
@@ -18,12 +45,17 @@ const { __ } = wp.i18n;
...
  */
-const VariantPanel = ( { title, variant, variants, onMouseDown } ) => {
+const VariantPanel = ( { index, blockId, postId, title, variant, variants, onMouseDown } ) => {
 	const { updateBlockAttributes } = useDispatch( 'core/block-editor' );
 
 	const defaultPercentage = 100 / variants.length;
 	const minPercentage = variants.length > 1 ? 0 : 100;
 
+	// Get A/B Test Block Id
+	const ab_post = useSelect( select => {
```
roborourke, line 55: I know in some of the code we have some snake case naming conventions - generally as a rule I always try to keep my code style consistent with the existing code in the file, so this variable name would become `abPost` for example.

inc/blocks/ab-test/edit.css
```diff
@@ -0,0 +1,13 @@
+button.ab-test-xb-preview__tab {
```
roborourke, line 1: The `button` part of the selector isn't strictly necessary. Also consider setting a default font-family, it will mirror whatever's on the site - you can look at how the font-family is set for the admin bar for that, think it might be via the `.wp-core-ui` class.

If you're having issues with styling and CSS specificity for something like this you could use `!important` - it's generally frowned on but this is a fair use-case for it.

inc/blocks/ab-test/edit.js
roborourke: Bonus space here:
```suggestion
						blockId={ attributes.clientId }
```

src/ab-test-preview.js
```diff
@@ -0,0 +1,57 @@
...
+
+		// set the class for the button
+		tab.className = 'ab-test-xb-preview__tab';
+
+		tab.addEventListener( 'click', function ( e ) {
+			// clone the correspnding data for the tab clicked
+			const variant = templates[i].content.cloneNode( true );
+
+			// remove the active class from any tab
+			xb.querySelectorAll( '.ab-test-xb-preview__tabs[data-client-tabs="' + xb.dataset.clientId + '"] .ab-test-xb-preview__tab' ).forEach( el => {
```
roborourke, line 33: You don't need the full `.ab-test-xb-preview__tabs[data-client-tabs="' + xb.dataset.clientId + '"]` selector  here, you can use a scoped query like so:

```suggestion
			tabContainer.querySelectorAll( '.ab-test-xb-preview__tab' ).forEach( el => {
```

The `data-client-tabs` attribute isn't needed.

inc/blocks/ab-test/register.php
```diff
@@ -113,6 +113,30 @@ function render_block( array $attributes, ?string $inner_content = '' ) : string
...
+		);
+		wp_enqueue_script(
+			'ab-test-preview',
+			Utils\get_asset_url( 'blocks/ab-test-preview.js' ),
+			[],
+			null
+		);
+
+		return sprintf(
+			'%s<div class="ab-test-xb-preview %s" data-client-id="%s" post-id="%s"><div class="ab-test-xb-preview__tabs" data-client-tabs="%3$s"></div><div class="ab-test-xb-preview__content" data-client-content="%3$s"></div></div>',
```
roborourke, line 131: The extra data attributes that repeat the client ID shouldn't be necessary, it can all be derived the scope you have in the `forEach()` loop over the blocks.

Additionally using `post-id` on it's own is invalid HTML, for custom attributes on regular HTML elements use data attributes e.g. `data-post-id`.

We use `post-id` as an attribute on the `<ab-test-block>` web component because we can define everything about a custom element, we can't define any new attributes for existing tags like `<div>`.

```suggestion
			'%s<div class="ab-test-xb-preview %s" data-client-id="%s" data-post-id="%s"><div class="ab-test-xb-preview__tabs"></div><div class="ab-test-xb-preview__content"></div></div>',
```

src/ab-test-preview.js
```diff
@@ -0,0 +1,57 @@
...
+	for ( let i = 0; i < templates.length; i++ ) {
+		// create a button element
+		const tab = document.createElement( 'button' );
+
+		// set the button inner html to the test title
+		tab.innerHTML = templates[i].dataset.title;
+
+		// if the title is empty the set the title based on the index
+		if ( templates[i].dataset.title === '' ) {
+			tab.innerHTML = 'Variant ' + getLetter( i );
```
roborourke, line 22: As far as possible code should be translatable, so you could use the `__()` function to do this. See other JS files for how to include functions from `wp.i18n`.

src/ab-test-preview.js
```diff
@@ -0,0 +1,57 @@
+import { getLetter } from '../src/utils';
+
+document.querySelectorAll( '.ab-test-xb-preview' ).forEach( xb => {
+
+	const xbPostId = xb.getAttribute( 'post-id' );
```
roborourke, line 5: When changing this attribute to be a `data` attribute will need to change how the value is read:
```suggestion
	const xbPostId = xb.dataset.postId;
```

Note that data attributes are automatically camel cased when reading from the `dataset` property, so `post-id` becomes `postId` etc...

src/ab-test-preview.js
```diff
@@ -0,0 +1,57 @@
...
+		} );
+
+		// append the created tab
+		tabContainer.appendChild( tab );
+	}
+
+	// determine if a specific tab should be clicked
+	// else click the first tab
+	if ( url_test ){
+		tabContainer.children[url_test[2]].click();
```
roborourke, line 52: We're ok in this instance but sometimes you have to be careful when using the `children` property of an element as even spaces between HTML tags are counted as child text nodes. The following would be extra safe but not needed here as we're creating the contents of this container directly.
```suggestion
		tabContainer.querySelectorAll( 'button' )[url_test[2]].click();
```

inc/blocks/ab-test/register.php
```diff
@@ -113,6 +113,30 @@ function render_block( array $attributes, ?string $inner_content = '' ) : string
...
+		wp_enqueue_style(
+			'altis-experiments-features-blocks-abtest',
+			plugins_url( 'inc/blocks/ab-test/edit.css', Analytics\ROOT_FILE ),
+			[],
+			null
+		);
+		wp_enqueue_script(
+			'ab-test-preview',
+			Utils\get_asset_url( 'blocks/ab-test-preview.js' ),
+			[],
```
roborourke, line 126: When using `wp.i18n` functions in a script you need to ensure that `wp-i18n` is in the the list of dependencies for the script when you enqueue it.

inc/blocks/ab-test/edit.css
```diff
@@ -0,0 +1,13 @@
...
+	background-color: #4667de;
```
roborourke, line 2: Basic theme styles might set other background properties like background-image so best to just completely clobber them using `background`:
```suggestion
	background: #4667de;
```

inc/blocks/ab-test/edit.css
```diff
@@ -0,0 +1,13 @@
...
+	padding: 0.25rem 0.5rem;
+	font-size: 0.75rem;
+	margin-right: 0.25rem;
```
roborourke, line 5: Similar here for margin:
```suggestion
	margin: 0 0.25rem 0 0;
```

inc/blocks/ab-test/edit.css
```diff
@@ -0,0 +1,13 @@
...
+}
+
+button.ab-test-xb-preview__tab.active,
+button.ab-test-xb-preview__tab:hover,
+button.ab-test-xb-preview__tab:focus {
+	background-color: #152a4e;
```
roborourke, line 11:
```suggestion
	background: #152a4e;
```

roborourke (changes requested):
Looking really good @robindevitt, couple of last tweaks to do I think. Can you add a screenshot of the front end tabs and the backend preview link to the PR comment too?

inc/blocks/ab-test/register.php
```diff
@@ -113,6 +113,29 @@ function render_block( array $attributes, ?string $inner_content = '' ) : string
...
+			null
+		);
+
+		return sprintf(
+			'%s<div class="ab-test-xb-preview %s" data-client-id="%s"><div class="ab-test-xb-preview__tabs" data-client-tabs="%3$s"></div><div class="ab-test-xb-preview__content" data-client-content="%3$s"></div></div>',
```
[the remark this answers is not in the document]
robindevitt: Will be removed in a newer commit

src/ab-test-preview.js
robindevitt: This was updated an changed to a different method

src/ab-test-preview.js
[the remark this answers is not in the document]
robindevitt: This should have been resolved in a newer commit

src/ab-test-preview.js
robindevitt: This is resolved through another method

inc/blocks/ab-test/register.php
robindevitt: The `post-id`/`data-post-id` is no longer needed with the changes made for the preview URL's so I have removed these now.

src/ab-test-preview.js
robindevitt: Thanks, this one slipped through. I initially went to do the translate but left it out on as my initial thoughts were that the user will have entered it in the desired language, however the "fallback" title slipped my mind.

robindevitt:
Tabs backend URL preview :
tabs-backend-preview

Tabs front end preview:

tabs-front-end

inc/blocks/ab-test/components/variant-panel.js
robindevitt: @roborourke Would the component file be located inside the `src/components` folder? Once it's working, do I then adjust the titles AB test code to use the same compnenet?

roborourke: Yeah I think that makes sense, and yes we'd want to use it in both places. At this point I'm thinking you could just do this in a follow up PR instead though. Lets try and get the core feature sorted today.

src/ab-test-preview.js
```diff
@@ -0,0 +1,59 @@
...
+			tabContent.appendChild( variant );
+		} );
+
+		// append the created tab
+		tabContainer.appendChild( tab );
+	}
+
+	// determine if a specific tab should be clicked
+	// else click the first tab
+	if ( url_test ){
```
roborourke, line 53:
```suggestion
	if ( url_test ) {
```

inc/blocks/ab-test/register.php
```diff
@@ -113,6 +113,32 @@ function render_block( array $attributes, ?string $inner_content = '' ) : string
 		$winner = $results['winner'] ?? false;
 	}
 
+	if ( is_preview() || isset( $_GET['set_test'] ) ) {
```
roborourke, line 116: I think this means I could use the `set_test` query string parameter even if I'm not logged in or able to edit the content, would be good to check permissions here before showing the preview mode.

roborourke (commented):
Alright - one item regarding a permissions check, some CSS linter errors and missing space to go then this is good. Nice work

inc/blocks/ab-test/register.php
robindevitt: I have added a check for if a user can edit posts and wrapped the preview check in the if statement. PR has been updated

robindevitt:
Just doing some tests and came across a bug when using the AB Test block in place of a widget. I'm not getting the post ID for the AB Block and there can't be a page ID either.

inc/blocks/ab-test/register.php
```diff
@@ -113,6 +113,34 @@ function render_block( array $attributes, ?string $inner_content = '' ) : string
 		$winner = $results['winner'] ?? false;
 	}
 
+	if ( current_user_can( 'edit_posts' ) ) {
+		if ( is_preview() || isset( $_GET['set_test'] ) ) {
```
roborourke, line 117: If we follow the principle of least privilege and we can check a precise post ID we should. This has the benefit of checking post type etc...
```suggestion
	if ( current_user_can( 'edit_post', $post->post_parent ?? 0 ) && ( is_preview() || isset( $_GET['set_test'] ) ) ) {
```

`$post->post_parent` is the ID of the post the XB was created on.

inc/blocks/ab-test/register.php
```diff
@@ -113,6 +113,34 @@ function render_block( array $attributes, ?string $inner_content = '' ) : string
...
+
+			return sprintf(
+				'%s<div class="ab-test-xb-preview %s" data-client-id="%s" data-post-id="%s"><div class="ab-test-xb-preview__tabs"></div><div class="ab-test-xb-preview__content"></div></div>',
+				$inner_content,
+				$class_name,
+				$client_id,
+				$post->ID ?? ''
+			);
+		}
+	};
```
roborourke, line 142: We don't normally put semicolons after if statements, it's not necessary here at any rate

roborourke:
> Just doing some tests and came across a bug when using the AB Test block in place of a widget. I'm not getting the post ID for the AB Block and there can't be a page ID either.

The XB post is only created on `save_post` at the moment I believe so that would make sense. If so then the bug affects personalised content too. We would need to create the XB post on the action that saves widgets too but I think that can be a separate issue & PR.

Ideally we'd only need to check `is_preview()` here really but would have to generate the query string args like `preview_id=<id>&preview_nonce=<nonce>&preview=true`.

inc/blocks/ab-test/register.php
robindevitt: Nasty and bad habit of mine!

Issue #298, opened 2021-12-13
A/B Test preview from a widget


robindevitt:
#### Issue
Currently you can't successfully preview an A/B Test block from within the widget area or customizer.

[Link to Slack thread](https://hmn.slack.com/archives/C5WSNGC9F/p1639404206213400) relating to the issue.

#### Accepted Criteria
- Successfully view an A/B Test Block from with the widget area
- Successfully view an A/B Test Block and edit the content within the customizer

roborourke:
Fixed by #306

Issue #299, opened 2021-12-14
Preview link as a component


robindevitt:
Currently we are reusing code to generate the preview link. This issue will create the preview link as a shared component that be reused.

[Initial suggestion for the change](https://github.com/humanmade/aws-analytics/pull/295#discussion_r766799430)

robindevitt:
Hey @roborourke 

On the [XB Review Widget](https://github.com/humanmade/aws-analytics/pull/300) we removed the preview links for A/B test blocks. 

Im not sure if that change impacts the need for making the preview link a component?

roborourke:
Yeah I think we can close this one for now. We might do more as at some point we should refactor everything to use typescript.

PR #300, opened 2021-12-14
XB preview widget

robindevitt:
Allow publishing and previewing of AB Test blocks from the widget screen

https://github.com/humanmade/aws-analytics/issues/298


inc/blocks/namespace.php
```diff
@@ -157,6 +158,84 @@ function on_save_post( int $post_ID, WP_Post $post, bool $update ) : void {
...
+		}
+
+		// Extract existing post if found.
+		$posts = wp_list_filter( $existing_posts->posts, [
+			'post_name' => $xb['attrs']['clientId'],
+		] );
+		$posts = array_values( $posts ); // Reset keys.
+
+		// Generate a default using the current post title and instance number in the content.
+		$default_title = sprintf( '(XB %s)', $index + 1 );
```
robindevitt, line 198: Im finding that there is only a single item so the  index is always going to be 1. 

Originally the code was : `$default_title = sprintf( '%s (XB %s)', $post->post_title, $index + 1 );`

However there is no `$post->post_title`, hence the change.

( I need to adjust the comment more appropriately for the title once a solution is possibly implemented here )

robindevitt:
This is currently a WIP but just facing a couple "issues".

1. Reference to this comment : https://github.com/humanmade/aws-analytics/pull/300#discussion_r769406647
2. The preview link may need to be adjusted as when creating a new A/B test block with variants as a widget, it is only published once the widget page is updated. The preview URL isn't updated until the page has reloaded, so initially once the page is updated the preview url contains an undefined value but once the page is refreshed it is then populated.

inc/blocks/namespace.php
roborourke: So each block counts as a widget? Can’t have multiple blocks in a widget? If so then it’s fine to omit that, you might even be able to fall back to or reference the widget title from the instance data

roborourke:
I think the preview link should be hidden until the XB post exists to avoid that issue

robindevitt:
> I think the preview link should be hidden until the XB post exists to avoid that issue

Should this be implement when creating the preview link as a component in [Issue : Preview link as a component](https://github.com/humanmade/aws-analytics/issues/299) ?

roborourke:
> Should this be implement when creating the preview link as a component in Issue : Preview link as a component ?

It does seem like a separate issue to this one. We might possibly want to change how the preview link for the AB test block works so it isn't dependent on having a parent post - it could use the block client ID attribute as you were doing initially -sorry!.

roborourke:
Actually saying that, yeah - makes sense to do any necessary preview link fixes in this PR to get this working. Although I'm actually now leaning towards not having the preview links for the AB test block - at least in the widget context.

You can preview a post but I don't think you can really easily preview widgets.

roborourke:
Per convo on Slack I think the best course of action is:

- Remove preview links for A/B test block
- Show preview output when either `is_preview()` or `is_customize_preview()` is true

robindevitt:
@roborourke 

> Remove preview links for A/B test block

Just to confirm, do we want to remove the preview links for all A/B test blocks or only on the widget screen? For consistency is it better to hide the preview link completely?  I was thinking we could do a check for the `postId` and  `abPost` which would allow the preview to still work on posts and pages but hidden on the widget screen.

robindevitt:
When adding the widget on the default widget screen everything appears as expected.

I started looking into the use of the widgets on the customizer screen today. When editing a widget it goes blank and the preview disappears. Secondly I noticed when saving the widgets it uses the action `save_post` and function `on_save_post` rather than the filter `widget_update_callback` and function `on_widgets_save` added for the widgets when viewing the widget page.

roborourke:
>> Remove preview links for A/B test block
>
> Just to confirm, we want to remove the preview links for consistency it's better to hide the preview link completely? Rather than only having it partially available?

I think for consistency yeah lets just remove it completely from the A/B testing block, sorry - was good to see it working but in this instance lets rely on the existing preview methods.

roborourke:
> I started looking into the use of the widgets on the customizer screen today. When editing a widget it goes blank and the preview disappears. Secondly I noticed when saving the widgets it uses the action save_post and function on_save_post rather than the filter widget_update_callback and function on_widgets_save added for the widgets when viewing the widget page.

This is interesting, What do you know about the post? Was just saving to post while you made changes? The customiser has 2 stages where it saves your changes as you go but then only applies them when you publish your changes so maybe that's why the post hook is firing - maybe the widget save hook is firing once you hit publish for the customiser changes.

robindevitt:
> This is interesting, What do you know about the post? Was just saving to post while you made changes? The customiser has 2 stages where it saves your changes as you go but then only applies them when you publish your changes so maybe that's why the post hook is firing - maybe the widget save hook is firing once you hit publish for the customiser changes.

Correct the `save_post` fires when publishing the customiser changes. I haven't found a hook for while the user is making changes.

Currently it fails on line - https://github.com/humanmade/aws-analytics/blob/71c1530d1eff215d0a060fae4b1a95a87c7a4f62/inc/blocks/namespace.php#L101
`return $xb['attrs']['clientId'];`

There is the `$post` data which contains the post info and could probably extract the `clientId` from there.

robindevitt:
When viewing the widget page, I have the following show up in Query Monitor: 
>wp_enqueue_script() was called incorrectly. "wp-editor" script should not be enqueued together with the new widgets editor (wp-edit-widgets or wp-customize-widgets). Please see Debugging in WordPress for more information. (This message was added in version 5.8.0.)

roborourke:
> wp_enqueue_script() was called incorrectly. "wp-editor" script should not be enqueued together with the new widgets editor (wp-edit-widgets or wp-customize-widgets). Please see Debugging in WordPress for more information. (This message was added in version 5.8.0.)

Yeah don't worry about this for now - I saw it too but so many existing blocks and things have the same issue there are many places we'd need to address that. Can you create an issue on the CMS module repo for it? Will need to do some quite extensive changes on multiple dependencies to fix it.

roborourke:
> There is the $post data which contains the post info and could probably extract the clientId from there.

Can you elaborate - it's a bit weird that as if it gets to this point it suggests to me that it has found some XBs in the `post_content` and those should have a clientId attribute.

What's the post type of the post being saved and what does the content look like? Maybe we need to skip processing that particular post type.

robindevitt:
This is a dumo on the `$post` variable 
```
(
	[ID] => 64
	[post_author] => 1
	[post_date] => 2021-12-20 14:59:25
	[post_date_gmt] => 2021-12-20 14:59:25
	[post_content] => {
		"widget_block[18]": {
			"value": {
				"raw_instance": {
					"content": "<!-- wp:altis/ab-test {\"clientId\":\"dd84c36f-d843-41b9-96c1-7ba5c094ce0e\"} -->\n<!-- wp:altis/ab-test-variant {\"fallback\":true,\"percentage\":100} -->\n<!-- wp:paragraph -->\n<p>Test from widget customiser again and again</p>\n<!-- /wp:paragraph -->\n<!-- /wp:altis/ab-test-variant -->\n<!-- /wp:altis/ab-test -->"
				}
			},
			"type": "option",
			"user_id": 1,
			"date_modified_gmt": "2021-12-20 14:59:25"
		}
	}
	[post_title] => 
	[post_excerpt] => 
	[post_status] => trash
	[comment_status] => closed
	[ping_status] => closed
	[post_password] => 
	[post_name] => baf2a1b3-21fb-48eb-b57d-c02f5c907f67
	[to_ping] => 
	[pinged] => 
	[post_modified] => 2021-12-20 14:59:25
	[post_modified_gmt] => 2021-12-20 14:59:25
	[post_content_filtered] => 
	[post_parent] => 0
	[guid] => urn:uuid:bdcb2294-8505-436c-b09e-75d819eacf53
	[menu_order] => 0
	[post_type] => customize_changeset
	[post_mime_type] => 
	[comment_count] => 0
	[filter] => raw
)
```

robindevitt:
When updating the customiser, this is a dump on the `$xbs` variable: 
```
(
     [0] => Array
         (
             [blockName] => altis/ab-test
             [attrs] => 
             [innerBlocks] => Array
                 (
                     [0] => Array
                         (
                             [blockName] => altis/ab-test-variant
                             [attrs] => 
                             [innerBlocks] => Array
                                 (
                                     [0] => Array
                                         (
                                             [blockName] => core/paragraph
                                             [attrs] => Array
                                                 (
                                                 )
 
                                             [innerBlocks] => Array
                                                 (
                                                 )
 
                                             [innerHTML] => \n<p>Test from widget customiser</p>\n
                                             [innerContent] => Array
                                                 (
                                                     [0] => \n<p>Test from widget customiser</p>\n
                                                 )
 
                                         )
 
                                 )
 
                             [innerHTML] => \n\n
                             [innerContent] => Array
                                 (
                                     [0] => \n
                                     [1] => 
                                     [2] => \n
                                 )
 
                         )
 
                 )
 
             [innerHTML] => \n\n
             [innerContent] => Array
                 (
                     [0] => \n
                     [1] => 
                     [2] => \n
                 )
 
         )
 
 )
```

roborourke:
Whoa! That's unexpected 🤔  So we definitely need some special handling for the `customize_changeset` post type. So is the `post_content` there a string of JSON? Or an array?

If that's just tracking changes until the customiser changeset is published then I think we just need to avoid further processing if it's changeset post type in the same way that we return early if the current post type is the XB post type.

robindevitt:
>Whoa! That's unexpected 🤔 So we definitely need some special handling for the customize_changeset post type. So is the `post_content` there a string of JSON? Or an array?

`post_content` is a string of JSON

inc/blocks/namespace.php
```diff
@@ -73,11 +73,18 @@ function on_save_post( int $post_ID, WP_Post $post, bool $update ) : void {
...
+		'customize_changeset'
+	);
+
+	if ( in_array( $post->post_type, $post_types ) ) {
 		return;
 	}
 
-	if ( $post->post_type === 'revision' ) {
+	$post_type_object = get_post_type_object( $post->post_type );
+	if( !$post_type_object->public || !$post_type_object->show_in_rest ){
```
roborourke: We want to process the post if _either_ `public` or `show_in_rest` are true. Currently this is saying:

> if `public` is false then return early, or if `show_in_rest` is false then return early

Whereas what we really need to say is

> if `public` is and false and `show_in_rest` is also false then return early

Because we're returning early the logic is reversed - it's a bit to get your head around at first when using early returns. If the code we wanted to execute was inside the `if` block then the `||` operator would make sense.

Remember to keep an eye on spacing and code style too.

```suggestion
	if( ! $post_type_object->public && ! $post_type_object->show_in_rest ) {
```

inc/blocks/ab-test/register.php
```diff
@@ -113,7 +113,7 @@ function render_block( array $attributes, ?string $inner_content = '' ) : string
 		$winner = $results['winner'] ?? false;
 	}
 
-	if ( current_user_can( 'edit_post', $post->post_parent ?? 0 ) && ( is_preview() || isset( $_GET['set_test'] ) ) ) {
+	if ( current_user_can( 'edit_post', $post->post_parent ?? 0 ) && ( is_preview() || is_customize_preview() ) ) {
```
roborourke, line 116: I'm thinking about it now and if we only preview on `is_preview()` or `is_customize_preview()` then the user will have appropriate privileges so we can drop the `current_user_can()` part - the XB post won't always have a parent post in the case of those added as widgets.
```suggestion
	if ( is_preview() || is_customize_preview() ) {
```

inc/blocks/namespace.php
```diff
@@ -73,11 +73,18 @@ function on_save_post( int $post_ID, WP_Post $post, bool $update ) : void {
 		return;
 	}
 
-	if ( $post->post_type === POST_TYPE ) {
+	$post_types = array(
+		POST_TYPE,
+		'revision',
+		'customize_changeset'
+	);
```
roborourke: Short array syntax:
```suggestion
	$post_types = [
		POST_TYPE,
		'revision',
		'customize_changeset'
	];
```
I'd probably also call the variable something more explanatory like `$excluded_post_types`, wouldn't hurt to add a filter for that too.

inc/blocks/namespace.php
```diff
@@ -72,11 +73,12 @@ function on_save_post( int $post_ID, WP_Post $post, bool $update ) : void {
...
 	}
 
-	if ( $post->post_type === POST_TYPE ) {
+	if ( in_array( $post->post_type, [ POST_TYPE, 'revision', 'customize_changeset' ], true ) ) {
 		return;
 	}
 
-	if ( $post->post_type === 'revision' ) {
+	$post_type_object = get_post_type_object( $post->post_type );
+	if ( ! $post_type_object->public || ! $post_type_object->show_in_rest ) {
```
roborourke, line 81: We only want to return early if both of these are false
```suggestion
	if ( ! $post_type_object->public && ! $post_type_object->show_in_rest ) {
```

inc/blocks/ab-test/components/variant-panel.js
```diff
@@ -92,15 +62,6 @@ const VariantPanel = ( { index, blockId, postId, title, variant, variants, onMou
 					} );
 				} }
 			/>
-			{ !! abPost && (
-				<Preview
-					href={ `/?p=${ postId }&set_test=test_xb_${ abPost.id }:${ index }` }
```
roborourke, line 97: Looks like we no longer need the `index`, `blockId` or `postId` props now.

inc/blocks/namespace.php
roborourke: This will always be "(XB 1)" which isn't a super helpful title.

If you update the action hook to pass the 3rd parameter which is the `WP_Widget` object, that has a `number` property that could be used in place of `$index + 1`, e.g. `$widget->number`. Don't forget to update how many arguments the hook receives too.

I would probably also check the block type, plus we know the context here so we can generate default titles like:

"Widget A/B Test (XB 1)"

Ideally in the JS we would have the same kind of fallback placeholder title if possible, you need to look at the context, maybe if `currentPost` is null?

src/ab-test-preview.js
```diff
@@ -12,16 +12,18 @@ document.querySelectorAll( '.ab-test-xb-preview' ).forEach( xb => {
 	const regex = new RegExp( `(utm_campaign|set_test)=test_xb_${ xbPostId }:(\\d+)`, 'i' );
 	const url_test = unescape( window.location.search ).match( regex );
 
-	for ( let i = 0; i < templates.length; i++ ) {
+	const total_templates = templates.length;
```
roborourke, line 15: Small note here - I know `url_test` came from some of my code but in JS we prefer to go for camelCase variable names these days just to match the way react likes to name things.

The main thing is always to try and maintain consistency with the existing code so use your judgment there.

roborourke (commented):
Few more comments sorry, it's looking good so keep at it :)

We definitely need to find a reasonable default title for the block and also ideally I'd like to avoid too much code duplication between how we handle saving a post and saving widgets.

inc/blocks/ab-test/register.php
robindevitt: As much as I had thought about dropping that... 

>Any visitor can add ?preview=true, ?preview=1, or ?preview=not_false to the URL, and is_preview() will return true.

I realise it's [mentioned in the comments](https://developer.wordpress.org/reference/functions/is_preview/#user-contributed-notes) and not the official docs so it may be worthwhile keeping it in place?

roborourke: Oh right, hmmm. Thing is if the XB has no parent post that'll never return true except for super admins. Maybe just change it back to the primitive capability `current_user_can( 'edit_posts' )` then.

inc/blocks/ab-test/edit.js
```diff
@@ -215,7 +215,7 @@ const Edit = ( {
 						blockId={ attributes.clientId }
 						className={ `variant-settings-${ variant.clientId }` }
 						index={ index }
-						postId={ currentPost.id }
+						postId={ currentPost.id ?? '' }
```
roborourke, line 218: Props can be removed here too

inc/blocks/namespace.php
```diff
@@ -157,6 +159,86 @@ function on_save_post( int $post_ID, WP_Post $post, bool $update ) : void {
...
+		}
+
+		// Extract existing post if found.
+		$posts = wp_list_filter( $existing_posts->posts, [
+			'post_name' => $xb['attrs']['clientId'],
+		] );
+		$posts = array_values( $posts ); // Reset keys.
+
+		// Generate a default using the current post title and instance number in the content.
+		$default_title = sprintf( 'Widget A/B Test (XB %s)', $widget->number );
```
roborourke, line 201: Sorry - I didn't mean that entirely literally, I was trying to point you in the right direction. The XB being processed here could be an A/B test or personalised content. You can check `$xb['blockName']` to see which type it is.

src/ab-test-preview.js
```diff
@@ -10,18 +10,20 @@ document.querySelectorAll( '.ab-test-xb-preview' ).forEach( xb => {
 	const tabContent = xb.querySelector( '.ab-test-xb-preview__content' );
 
 	const regex = new RegExp( `(utm_campaign|set_test)=test_xb_${ xbPostId }:(\\d+)`, 'i' );
-	const url_test = unescape( window.location.search ).match( regex );
+	const urlTest = unescape( window.location.search ).match( regex );
 
-	for ( let i = 0; i < templates.length; i++ ) {
+	const totalTemplates = templates.length;
```
roborourke, line 15: Nice optimisation here 👍

inc/blocks/namespace.php
```diff
@@ -157,6 +159,86 @@ function on_save_post( int $post_ID, WP_Post $post, bool $update ) : void {
 	}
 }
 
+/**
+ * Synchronise any XBs on the widget page with a shadow post type.
+ *
+ * @param instance $instance Widget Object.
+ * @param new_instance $new_instance Widget Object.
+ * @param old_instance $old_instance Widget Object
+ * @param widget $widget Current Widget Instance
```
roborourke, line 168: This isn't proper docblock syntax, not sure if it was generated but you should know what the components are and how they should be written, see this guide for an overview: https://www.open-emr.org/wiki/index.php/How_to_Document_Your_Code_Properly

For the `@param` annotation you'll see in other examples in the file that it's `@param <type> <variable name> <description>`.

- The `type` is any valid PHP variable type such as `array`, `bool`, `string` or a class name etc...
- The `description` should end in a full stop and be a bit more descriptive whenever possible

The same goes for the return type "object", although valid it is not what is returned by this filter, it's actually an `array`. See the filter docs: https://developer.wordpress.org/reference/hooks/widget_update_callback/

inc/blocks/namespace.php
```diff
@@ -157,6 +159,86 @@ function on_save_post( int $post_ID, WP_Post $post, bool $update ) : void {
...
+ * @return object
+ */
+function on_widgets_save( $instance, $new_instance, $old_instance, $widget ) {
+
+	// Scan for XBs in the post content.
+	$blocks = parse_blocks( $new_instance['content'] );
```
roborourke, line 174: I think you can just check `$instance['content']` here as `$instance` has already been through the widget class's `update` method.

inc/blocks/namespace.php
```diff
@@ -157,6 +159,86 @@ function on_save_post( int $post_ID, WP_Post $post, bool $update ) : void {
...
+		/**
+		 * Allow further processing after the XB post is created or updated.
+		 *
+		 * @param int $post_id The XB post ID.
+		 * @param array $xb The unserialised block data.
+		 */
+		do_action( 'altis.analytics.blocks.save_post', $xb_post_id, $xb );
+	}
+
+	return $new_instance;
```
roborourke, line 239: This can just return `$instance` as we're doing no modifications to it. A filter should generally return the first argument passed into it, or a modified version of that first argument.

inc/blocks/namespace.php
```diff
@@ -157,6 +159,86 @@ function on_save_post( int $post_ID, WP_Post $post, bool $update ) : void {
...
+		} else {
+			// Update existing post.
+			$xb_post_id = $posts[0]->ID;
+			wp_update_post( [
+				'ID' => $posts[0]->ID,
+				'post_content' => serialize_block( $xb ),
+				'post_title' => $xb['attrs']['title'] ?: $default_title,
+			] );
+		}
+
```
roborourke, line 229: Lets set some post meta so we know this XB is a widget, maybe with the key `_xb_is_widget`

inc/blocks/namespace.php
```diff
@@ -157,6 +159,86 @@ function on_save_post( int $post_ID, WP_Post $post, bool $update ) : void {
...
+ * Synchronise any XBs on the widget page with a shadow post type.
+ *
+ * @param instance $instance Widget Object.
+ * @param new_instance $new_instance Widget Object.
+ * @param old_instance $old_instance Widget Object
+ * @param widget $widget Current Widget Instance
+ * @return object
+ */
+function on_widgets_save( $instance, $new_instance, $old_instance, $widget ) {
+
```
roborourke, line 172: Occurred to me that we're relying on `$instance` or `$new_instance` having the `content` key, for backwards compatibility can you add a check to return early is `! isset( $instance['content'] )`? Not sure if there's a more robust way to check if this is a block widget or not.

inc/blocks/namespace.php
robindevitt: Should it not be set here? https://github.com/humanmade/aws-analytics/blob/972aea11d28fffb3dc1fea3cf4982c0dc6dea86b/inc/blocks/namespace.php#L139-L140

roborourke: Sorry I thought I'd mentioned rebasing this PR against `master` - if you look at the latest code you'll see it has changed a bit.

If you're not familiar with it rebasing in git means to replay all of your commits over another branch - most often you'd do this to make sure your branch will merge cleanly and that you have the latest code. After making sure your primary branch is up to date, from your working branch you would run `git rebase master`.

You may run into conflicts but the output should tell you how to go about resolving these.

inc/blocks/ab-test/edit.js
```diff
@@ -191,7 +191,7 @@ const Edit = ( {
 					{ currentPost.type !== 'xb' && (
 						<TextControl
 							label={ <strong>{ __( 'Block Title', 'altis-analytics' ) }</strong> }
-							placeholder={ sprintf( __( '%s (XB %d)', 'altis-analytics' ), currentPost.title, instance.current ) }
+							placeholder={ sprintf( __( '%s %d', 'altis-analytics' ), ( currentPost.title ? currentPost.title : __( 'A/B Test', 'altis-analytics' )  ), instance.current ) }
```
roborourke, line 194: You could use the `||` shorthand instead of a ternary here e.g.
```suggestion
							placeholder={ sprintf( __( '%s %d', 'altis-analytics' ), ( currentPost.title || __( 'A/B Test', 'altis-analytics' )  ), instance.current ) }
```

If the thing you are testing is the value you want to use when it's truthy then you can use the `thing || 'default'` syntax to set it. It's just a bit more compact.

inc/blocks/namespace.php
```diff
@@ -157,6 +160,90 @@ function on_save_post( int $post_ID, WP_Post $post, bool $update ) : void {
...
+		}
+
+		// Extract existing post if found.
+		$posts = wp_list_filter( $existing_posts->posts, [
+			'post_name' => $xb['attrs']['clientId'],
+		] );
+		$posts = array_values( $posts ); // Reset keys.
+
+		// Generate a default using the current post title and instance number in the content.
+		$default_title = sprintf( __( 'Widget %s %d', 'altis-analytics' ), ucfirst( Blocks\get_block_type( $xb['attrs']['clientId'] ) ), $widget->number );
```
roborourke, line 206: The `get_block_type()` function is in the current namespace so you don't need to use the `Blocks` namespace prefix.
```suggestion
		$default_title = sprintf( __( 'Widget %s %d', 'altis-analytics' ), ucfirst( get_block_type( $xb['attrs']['clientId'] ) ), $widget->number );
```

What are the outputs from this function and what are the results when using `ucfirst`? Would it look ok to end users?

For example see the following: https://github.com/humanmade/aws-analytics/blob/master/inc/dashboard/namespace.php#L216-L230

If you rebase against `master` you could add a `get_block_type_name()` function perhaps or refactor `get_block_type()` to add an extra parameter to called "output" like `get_block_type( $client_id, 'name' )` for example.

inc/blocks/namespace.php
```diff
@@ -157,6 +160,90 @@ function on_save_post( int $post_ID, WP_Post $post, bool $update ) : void {
...
+ */
+function on_widgets_save( $instance, $new_instance, $old_instance, $widget ) {
+
+	if ( ! isset( $instance['content'] ) ) {
+		return $instance;
+	}
+
+	// Scan for XBs in the post content.
+	$blocks = parse_blocks( $instance['content'] );
+	$xbs = find_xbs( $blocks );
```
roborourke, line 180: In the other function there were other early returns if no XB's were found, if we can avoid doing unnecessary DB queries then we should.

roborourke (changes requested):

inc/blocks/ab-test/edit.js
robindevitt: Thank you, am I also double translating the text with the `__( )` ?

inc/blocks/namespace.php
robindevitt: The `ucffirst` was making the title render as "Widget Personalization 22" which I suspect I can remove a and leave it as "Widget personalization 22" ?

roborourke:
Hmm, seems like there was maybe some issue with the rebase here? It's showing all the updates since this branch was created even though most of them exist on `master` already so it's hard to review. Did you update your master branch locally first?

inc/blocks/ab-test/edit.js
roborourke: Good point, yeah you don't need to translate the string "%s %d" any more.

robindevitt:
> Hmm, seems like there was maybe some issue with the rebase here? It's showing all the updates since this branch was created even though most of them exist on master already so it's hard to review. Did you update your master branch locally first?

Yup, I update the local master branch first and then did the rebase.

Exact steps were:
- Switched to `master`
- `git pull`
- switched back to `xb-preview-widget`
- `git rebase master`

Have I missed something else?

roborourke:
No that sounds right, strange. Will have to google it

robindevitt:
> We definitely need to find a reasonable default title for the block and also ideally I'd like to avoid too much code duplication between how we handle saving a post and saving widgets.

This is the final "big" change I am working on getting done.

roborourke:
I think just try doing a `git merge master` instead, I think in this case it’s ok as it’s a new major release we’re working on

roborourke:
Ah I might know what happened, after the first preview related PR got merged, did you switch back to master and update it before making the new branch or was it branched from the previous PR?

robindevitt:
> Ah I might know what happened, after the first preview related PR got merged, did you switch back to master and update it before making the new branch or was it branched from the previous PR?

I think that's what may have happened that I hadn't updated/pulled the main branch before creating the new branch? Anything I can do at the moment to resolve the issue or is it done and I just need to be more aware going forward?

roborourke:
Yeah just try merging in the master branch at this stage, per https://github.com/humanmade/aws-analytics/pull/300#issuecomment-1000391284

robindevitt:
I have checked the `master` branch and it's up to date.

When I checkout the `xb-preview-widget` branch and try `git merge master` it says that its "Already up to date."

roborourke:
And you’ve pushed 🤔 I don’t know then that’s super weird. You could try closing this PR then renaming your local branch and making a new PR from that. All very confusing tbh

robindevitt:
Closed this PR and created a new branch to try resolve an issue from not pulling before creating a new branch.

New PR : #306 
New Branch : [xb-preview-widget-new](https://github.com/humanmade/aws-analytics/tree/xb-preview-widget-new)

PR #301, opened 2021-12-15
Refactor experiements framework for extensibility

shadyvb:
This refactors both the PHP and JS code associated with registering experiments, to allow decoupling the commonly used code from the existing title experiment code, to make it easier for new experiment types to be registered with minimal code duplication.

PHP:
- Optimized the code a bit to DRY the script registration and post type support bits.

JS:
- Created `registerExperiment` function to register a new experiment type, which uses WordPress hooks to minimize the code duplication as possible.
- Added the featured image experiment to test the code duplication.

Open issues:
- ~Need a way to remove a variant, because titles didn't have a visual element to do so, used keyboard keys.~


inc/experiments/namespace.php
```diff
@@ -362,8 +372,19 @@ function register_post_ab_test( string $test_id, array $options ) {
...
 			'page',
 		],
+		'editor_scripts' => [],
 	] );
 
+	// Handle post type support filtering and registration.
+	$options['post_types'] = apply_filters( "altis.experiments.features.{ $test_id }.post_types", array_filter( (array) $options['post_types'] ) );
+	foreach ( $options['post_types'] as $post_type ) {
+		add_post_type_support( $post_type, "altis.experiments.{ $test_id }" );
+	}
```
roborourke, line 382: I think this reverses the logic I had. The results of `get_post_types_by_support( "altis.experiments.{ $test_id }" )` would need to be merged into `$options['post_types']` as that value is checked in other places like the background task handler.

inc/experiments/namespace.php
```diff
@@ -394,6 +415,73 @@ function register_post_ab_test( string $test_id, array $options ) {
...
+		return;
+	}
+
+	foreach ( $post_ab_tests as $test_id => $test ) {
+		if ( empty( $test['editor_scripts'] ) ) {
+			continue;
+		}
+
+		// Get supported post types.
+		$supported_post_types = get_post_types_by_support( "altis.experiments.{ $test_id }" );
```
roborourke, line 455: Should be read from `$test['post_types']` instead I think.

inc/experiments/namespace.php
```diff
@@ -394,6 +415,73 @@ function register_post_ab_test( string $test_id, array $options ) {
...
+ * Register block editor scripts requird for registered tests.
+ *
+ * @param string $hook Current screen hook.
+ *
+ * @return void
+ */
+function enqueue_experiments_editor_scripts( string $hook ) : void {
+	global $post_ab_tests;
+
+	$register_experiments_framework = apply_filters( 'altis.experiments.frontend.framework.enabled', false );
```
roborourke, line 428: Missing docblock, and I'm not sure why this is needed?

src/experiments/experiment.js
```diff
@@ -0,0 +1,17 @@
...
+ * @param {int} experiment.id Experiement test ID.
+ * @param {string} experiment.title Experiement test title.
+ * @param {string} experiment.singleTitle Experiement test single entry title.
+ * @param {React.ReactNode} experiment.component Experiement test panel field component.
+ * @param {object} experiment.dispatcher Experiement test panel dispatch handler.
+ * @param {object} experiment.selector Experiement test panel select handler.
+ */
+export const registerExperiment = experiment => {
+	// Register the panel for the experiement.
+	addFilter( 'altis.experiments.experiments', `altis.experiments.${ experiment.id }`, experiments => [ ...experiments, experiment ] );
```
roborourke, line 16: I think as this follows the pattern of having to register editor blocks on both the backend and frontend maybe we can use the `block.json` pattern 🤔 I digress anyway, that's more for implementors to decide.

Anyway - I think what would be beneficial here is to validate that the experiment being registered is also registered on the backend, otherwise it's not going to do anything and the UI would just result in a ton of API errors.

Maybe `registerExperiment()` can follow the same conventions as `registerBlockType()`, `register_block_type()` and `register_post_ab_test()` e.g. the arguments are `func( id/name, settings/options )`.

Having the test ID as the first argument makes it easier to validate.

roborourke (commented):
This is really impressive! There's two big things that I think need to be addressed and we should be good to go - comments below.

It'd be great to get this new API documented as well - I've kind of been updating docs here and then copy/pasting them to the Altis docs then adding a few more examples.

src/experiments/experiment.js
shadyvb: Hmm, things is that the experiment ID is passed down with the experiment object down to all the components, otherwise I'll pass down the ID separately, OR merge the ID back into the object within the filter call here.

```js
/**
 * Register a new experiment type.
 *
 * @param {int} id Experiement test ID.
 * @param {object} options Experiement options.
 * @param {string} options.title Experiement test title.
 * @param {string} options.singleTitle Experiement test single entry title.
 * @param {React.ReactNode} options.component Experiement test panel field component.
 * @param {object} options.dispatcher Experiement test panel dispatch handler.
 * @param {object} options.selector Experiement test panel select handler.
 */
export const registerExperiment = ( id, options ) => {
	// Register the panel for the experiement.
	addFilter( 'altis.experiments.experiments', `altis.experiments.${ id }`, experiments => ( [
		...experiments,
		{
			id,
			...options,
		},
	] ) );
};
```

How would the validation work in this case ? Within the editor plugin code ? And how would passing the ID separately facilitate it ?

roborourke: Yeah you could merge it in to the object or have an object that acts as a registry, it doesn’t necessarily need to use a filter here looking at it. I think I suggested a filter thinking we’d have a default control and you just filter the variant input control.

Regarding validation you could serialise the registered AB tests when enqueueing the framework and cross reference that. I was expecting to share some data from the server side like the title / label etc

shadyvb: Hmmm, I think with the structure and generic fields we have, we can probably register stuff in PHP, pass that to the editor, and use the frontend to only _override_ default options for a specific experiment, that way we take care of validation, and minimize the duplication between PHP and JS code.

We can then pass down the experiment ID only, and grab the experiment object from the registry without having to pass down the full object everywhere.

re filters, I had a bunch of filters earlier, but they trimmed down to one, as I now pass the experiment option itself, so I thought the same about using a registry object, probably facilitate by the PHP<>JS communication just mentioned above.

shadyvb: Latest refactor makes the JS hooks completely optional, as you can create an experiment with an arbitrary value, and use that on the front-end as you please. We just need to document this, and the APIs that can be used for custom handling on front-end.

I've removed all hooks, and only used them to override test settings, which will be the same API we'll provide to custo integrations.

inc/experiments/namespace.php
roborourke: With the filter I wasn't really thinking of it as the core way of modifying a test at the time necessarily, I should've thought about that more and made a better implementation at the time. It's not strictly a filter that should be used anyway - that's up to us to decide, and I agree having both methods makes things confusing. I'm leaning towards getting rid of the filter just for the post types setting.

Previously if I'd used `add_post_type_support( 'type', 'altis.experiments.titles' )` it would have worked to add support for the titles test to that post type. That's what we're losing here at the moment with this change that I wanted to call out.

Ultimately the outcome I think we want is that it should be easy to add or remove support for a given test, and there should be a single canonical source of truth for getting supported post types, one of:

- `get_post_types_by_support( 'altis.experiments.<test>' )`
- `$post_ab_tests[ <test> ]['post_types']`

If we go with the post type support approach then we'll need to use `get_post_types_by_support( 'altis.experiments.<test>' )` as the way to get the array of post types. So we'd need to update any references to `$post_ab_tests[ <test> ]['post_types']`. What do you think?

Regarding filters in WP there is a pattern of having a filter or action that fires when registering something, like `register_post_type_args` and `registered_post_type` that might be a better pattern to follow than just having a filter for the post types option here, makes it much more pluggable.

src/experiments/data/registry.js
```diff
@@ -0,0 +1,126 @@
...
+ */
+export default class TestsRegistry {
+	tests = {};
+
+	/**
+	 * Initiatlize the class, transforming passed options to usable test objects.
+	 *
+	 * @param {object} tests Initial tests object to see the registry.
+	 */
+	constructor( tests ) {
```
roborourke, line 17: You're going to hate me for this but I'm not a fan of this approach to creating the registry.

- All post A/B tests will be registered in the sidebar UI - we explicitly don't want this for the A/B testing block so it should be opt in
- Having to create / expose a custom React component on the `window` object is a bit clunky, why not allow it to be passed when registering a test on the client side
- Most registry implementations use a `register` type method rather than updating existing items, this doesn't feel idiomatically similar to other existing APIs in WordPress

I preferred the opt-in approach with the `registerExperiment` function. It matched the way blocks are registered on the server and client side and meant it would be really straightforward to provide a custom component etc... I had in mind something along these lines:

```js
Altis.Analytics.Experiments.registerExperiment = ( id, options ) => {
   if ( ! Altis.Analytics.Experiments.PostABTests[ id ] ) {
     throw new Error( 'Test must be registered server side ... etc...' );
   }

   // Validate options...

   Altis.Analytics.Experiments.PostABTests[ id ] = {
     ...Altis.Analytics.Experiments.PostABTests[ id ],
     ...options,
     showUI: true,
   };
}
```

then to use it:

```js
const { registerExperiment } = Altis.Analytics.Experiments;

registerExperiment( 'titles', {
   component: TextInput,
   dispatcher( ( dispatch, values ) => { ... } ),
   ...
} );
```

Then check which AB tests have a component or the example flag like `showUI` or maybe `showInSidebar` to build the plugin sidebar.

It's a rough example but I think that kind of thing would to be a lot easier to work with than passing `input_type` from the backend or exposing react components on `window`.

Could still use a hook to pass a more contained registry object through if you prefer, doesn't have to be on the Altis window object.

roborourke (commented):
Feel like I'm giving you the runaround a bit here, let me know if you want to chat through any of my comments and I hope they're constructive. Really appreciate all the work you're putting in here.

inc/experiments/namespace.php
shadyvb: Yeah, I agree that using `get_post_types_by_support()` and getting rid of the filter makes the most sense, and about adding an action after registration. 👍

src/experiments/data/registry.js
roborourke: > I still prefer to use input_type to make the front-end registration optional. Maybe even go for field registration, which would make it even more extensible/reusable.

I'm not really convinced about setting this server side still, I don't want to have to maintain a bunch of built in field types when it's highly likely a custom dispatcher & selector would be needed anyway. We'd potentially be getting into passing all sorts of other options for a dropdown or field type that needs extra config, It'll get complicated quickly and it's more documentation to write and maintain.

I think the client side aspects will be easier to deal with purely on the client side and let implementors figure out what they need.

> I tried to get out of using globals for the test definitions object, but it was either a global ( eg: a registry of some sort / a global var ), or a filter, can also go for a context, something I hadn't thought of until now, and it makes quite more sense!

I'd probably leave a refactor like that for a later time tbh, there are few different ways/places that global data is used and was just simpler to understand and work with initially, especially once I had to start splitting out files so they could be seaprately enqueued. Globals aren't strictly a bad thing, they're easy to understand and can be useful - thinking of `wp.apiFetch`, `wp.i18n`, `wp.components.Thing` etc...

> I had some hard time following the terminology, I had started using experiment, then test, then abtest, then postabtest, trying to differentiate between the current test object that has the results and is being used everywhere, to avoid conflicts / confusion. We may need some refactoring there to make it less confusing. It'd be great if we can align now on all the terms and what they mean.

Yeah again, there was a fair bit of code churn and figuring things out as I went when I originally wrote it. "Post A/B Test" was named because a test was specifically associated with a post in the db. I originally wrote it so you could register different data storage and retrieval functions with a test but it was over engineered.

We can fix this in a later update I think for a v5. If it helps the following things are true:

- Experiments is a blanket term that gives us flexibility for different types of tests we add in future
- An A/B test is one type of experiment, a Multivariate test is another
- A test could be associated with a post, or potentially with the site or other entity (will see if the need for that actually comes up)

Hopefully that clears it up a bit.

In terms of how to proceed:

> - Use a `show_ui` prop to decide which experiments to add to the sidebar, during registration of the test/experiment.

✅  can use this to filter them out server side.

> - Use a context to pass down the JSON object, instead of the current kinda-registry approach. And to register custom components as well.

I don't fully get what advantage the context has, does it just make it more testable in isolation? I guess you'd do something like `useContext( 'experiments' )`? Can we add the context provider high enough or would we have to create it twice, like for the plugin icon and the plugin sidebar?

Again I wouldn't really go with the `input_type` option on the backend myself, I think there's always going to be some client side code to handle the dispatcher and selector and that's going to be specific per test rather than per field. Registering field types is getting a bit abstract in this instance as we're not making a form system.

Whichever way you go, as long as defining the field component for a test is easy and intuitive that's the main thing, a filter or function call to modify the test object would be simplest IMO. I'll leave it in your hands.

> - Use a JS event ( or a WP action/filter ) during panel registration to allow 3rd party code to adjust/override values.

✅ 

> - Use a PHP WP action at the end of register_post_ab_test() to allow 3rd party code to filter the registered test.

✅ 

> - Get rid of the post types support filter and use WP post type support functions all the way.

✅

shadyvb: The data inheritance here is perfect for React context, specially as with the `withTestData` HOC, it all becomes quite seamless.

46482b5 does the following:

- Use React Context to manage data inheritance, within `withTestData`, no prop passing anymore, each panel gets its own `abTest` prop if wrapped with `withTestData`, with available exported `context` object for any component ( eg the input ) usable with `useContext()` if needed.
- Use `show_ui` prop to skip tests with no sidebar UI.
- Remove `input_type` in favor of front-end explicit registration.
- Added a filter to manage registered test options before any handling, and an action to fire anything after the test is registered.
- Moved the REST API field registration to a separate hook on `rest_api_init` so new filters/actions (from the above) can action changes (like post type support) and have that accounted for there.
- Moved to use `get_post_type_by_support` all over.

src/experiments/data/index.js
```diff
@@ -0,0 +1,33 @@
...
+
+	for ( let id in input ) {
+		const test = input[ id ];
+
+		/* eslint-disable jsdoc/require-jsdoc */
+		tests[ id ] = applyFilters( `altis.experiments.sidebar.test.${ id }`, {
+			id,
+			title: test.label,
+			singleTitle: test.singular_label,
+			component: () => ( <></> ),
```
roborourke, line 20: Need to import React to this file unless this can default to `null`, if it's `null` we just wouldn't show the UI . Maybe we can infer a default control type from the variants field type setting?

roborourke (approved):
Nice one @shadyvb !

There are a couple linter errors, most important is importing react in the data HoC file.

Would it be possible to extend the readme a bit with the example for titles or images perhaps?

src/experiments/data/index.js
shadyvb: Yeah, I had actually wanted to use the default text input type. Good spot!

shadyvb: FYI the rule about having to import React is outdated, see https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html , might need to update our rules.

shadyvb:
This seems to work, so will merge, and work on two pending issues:
1. Updating the docs.
2. Addressing the issue where a post needs to have a featured image set before it can use the relative experiment.

So will merge this one here and build on top of it for the two issues above. cc @roborourke

PR #306, opened 2021-12-29
Xb preview widget new

robindevitt:
Continuation from #300 and [this suggested comment](https://github.com/humanmade/aws-analytics/pull/300#issuecomment-1002091851).

Allow publishing and previewing of AB Test blocks from the widget screen

#298


inc/blocks/namespace.php
```diff
@@ -131,7 +181,7 @@ function on_save_post( int $post_ID, WP_Post $post, bool $update ) : void {
 				'post_name' => $xb['attrs']['clientId'],
 				'post_author' => get_current_user_id(),
 				'post_title' => $xb['attrs']['title'] ?: $default_title,
-				'post_parent' => $post_ID,
+				'post_parent' => $post_ID ?: '',
```
roborourke, line 184: The field type in the database is an integer so if there's no parent use `null` or `0` as the fallback. This is why setting a default value in the parameters is sometimes useful.

inc/blocks/namespace.php
```diff
@@ -93,6 +95,49 @@ function on_save_post( int $post_ID, WP_Post $post, bool $update ) : void {
...
+/**
+ * Update or create new XB's
+ *
+ * @param int $post_ID Post ID if applicable in use case, currently no post ID for widgets.
+ * @param array $xbs XB's on page or as widgets.
+ * @param string $post_title Page/Post title for XB Post.
+ * @param int $widget_number XB as a widget needs a number as the index doesn't increment.
+ * @param string $xb_meta_key Meta Key for updating the post_meta.
+ */
+function update_create_xb( $post_ID, $xbs, $post_title, $widget_number, $xb_meta_key ) {
```
roborourke, line 140: When you're writing a function try to consider which parameters are always required, which can have default values, and which are optional, then aim to put them in that order. You should also aim for some degree of making things generic when abstracting functionality. In this case I'd do something like:

```suggestion
function update_create_xb( array $xbs, string $default_title = '', ?int $item_id = null, string $context = 'post' ) {
```

Also note the type hinting - it forces a function to be used correctly and reduces the capacity for errors and creating bugs. The question mark before a type e.g. `?int` means that it's a nullable type, so the value can be `null` or an integer for example.

inc/blocks/namespace.php
```diff
@@ -116,8 +161,13 @@ function on_save_post( int $post_ID, WP_Post $post, bool $update ) : void {
 		] );
 		$posts = array_values( $posts ); // Reset keys.
 
-		// Generate a default using the current post title and instance number in the content.
-		$default_title = sprintf( '%s (XB %s)', $post->post_title, $index + 1 );
+		if ( $post_title ) {
```
roborourke, line 164: If we pass the context and a default title we can probably make this logic a bit more generic.

roborourke (changes requested):
Some crucial things that need changing as we'll be breaking some existing functionality otherwise.

src/ab-test-preview.js
```diff
@@ -2,58 +2,79 @@ import { getLetter } from '../src/utils';
...
+// Create the tabs for previewing
+createTabbedPreviews();
 
+// When making changes in the customizer
+wp.customize.selectiveRefresh.bind( 'sidebar-updated', function () {
+	// clear out any exsisting tabs
+	document.querySelectorAll( '.ab-test-xb-preview' ).forEach( xb => {
+		const tabContainer = xb.querySelector( '.ab-test-xb-preview__tabs' );
+		tabContainer.innerHTML = '';
+	} );
```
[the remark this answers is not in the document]
robindevitt: When making a chance in the customizer it would add the tabs to the exsisting tabs, so essentially duplicate them and give me something like this:
Screenshot 2022-01-03 at 15 25 31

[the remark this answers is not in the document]
roborourke: Right but you could check in the function if they're already there rather than resetting.

inc/blocks/namespace.php
```diff
@@ -140,13 +190,13 @@ function on_save_post( int $post_ID, WP_Post $post, bool $update ) : void {
...
-				'post_parent' => $post_ID,
+				'post_parent' => $post_ID ?: '',
 			] );
 		}
 
-		// Store XB type, ensure back compat with existing XBs also.
-		update_post_meta( $xb_post_id, '_xb_type', $xb['blockName'] );
-		update_post_meta( $xb_post_id, '_xb_type_' . sanitize_key( $xb['blockName'] ), true );
+		// Store XB type.
+		update_post_meta( $xb_post_id, $xb_meta_key, $xb['blockName'] );
```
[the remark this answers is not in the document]
robindevitt: @roborourke I'm a little unsure on this comment and what I should be achieving here. I feel it's probably simple and that I am over thinking it?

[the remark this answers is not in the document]
roborourke: Yeah maybe overthinking a bit :) When I originally suggested adding some meta data to indicate that the XB was created as a widget I didn't mean you should change the existing post meta being saved. What was the thought process behind that change? Code reuse?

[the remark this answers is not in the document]
robindevitt: I may not be 100% here but I think I have a better idea now, it should be added with the the `wp_insert_post` rather than with the `update_post_meta`.

I didn't think we wanted to keep both the `xb_type` post meta, but I suspect the way I had gone with then there would be too many variants. Where as now we could query the post meta by `_xb_type` or the `_xb_content`?

robindevitt:
Hey @roborourke 
I think I'm almost there and I have published this now and requested a review when you ready.

src/ab-test-preview.js
```diff
@@ -2,58 +2,79 @@ import { getLetter } from '../src/utils';
...
+	document.querySelectorAll( '.ab-test-xb-preview' ).forEach( xb => {
 
-	const xbPostId = xb.dataset.postId;
-	const templates = document.querySelectorAll( 'template[data-parent-id="' + xb.dataset.clientId + '"]' );
-	const tabContainer = xb.querySelector( '.ab-test-xb-preview__tabs' );
-	const tabContent = xb.querySelector( '.ab-test-xb-preview__content' );
+		const xbPostId = xb.dataset.postId;
+		const templates = document.querySelectorAll( 'template[data-parent-id="' + xb.dataset.clientId + '"]' );
+		const tabContainer = xb.querySelector( '.ab-test-xb-preview__tabs' );
+		const tabContent = xb.querySelector( '.ab-test-xb-preview__content' );
```
roborourke, line 14: I can't see where we check if the tabs are already set up, I was thinking something like
```suggestion
		const tabContent = xb.querySelector( '.ab-test-xb-preview__content' );

		// Return if we've processed this XB already.
		if ( tabContainer.children.length > 0 ) {
			return;
		}
```

inc/blocks/namespace.php
```diff
@@ -382,6 +431,23 @@ function get_block_type( $post ) : string {
 	return str_replace( 'altis/', '', $type );
 }
 
+/**
+ * Readable block name based on the XB name.
+ *
+ * @param string $block_name Name of the XB.
+ * @return string
+ */
+function get_block_name( $block_name ) {
```
roborourke, line 440: I'd make this more specific as to what it's doing, might be a bit confusing if you're passing `$block_name` to the function, so perhaps `get_block_type_label()`.
```suggestion
function get_block_type_label( string $block_name ) : string {
```

Note adding the type hints too, always good practice for a function like this one.

Would be good to update the function in the dashboard namespace too to use this one so we only need to update a single function in future.

inc/blocks/namespace.php
```diff
@@ -93,6 +95,47 @@ function on_save_post( int $post_ID, WP_Post $post, bool $update ) : void {
...
+		update_create_xb( $xbs, 'Widget', null, $widget->number, 'widget' );
+	}
+
+	return $instance;
+}
+
+/**
+ * Update or create new XB's
+ *
+ * @param array $xbs XB's on page or as widgets.
```
roborourke, line 132: Just being pedantic here sorry, `'s` denotes ownership so doesn't make sense here.
```suggestion
 * Update or create new XB shadow posts.
 *
 * @param array $xbs XBs on page or as widgets.
```

inc/blocks/namespace.php
```diff
@@ -93,6 +95,47 @@ function on_save_post( int $post_ID, WP_Post $post, bool $update ) : void {
...
+		return $instance;
+	}
+
+	// Scan for XBs in the post content.
+	$blocks = parse_blocks( $instance['content'] );
+	$xbs = find_xbs( $blocks );
+
+	// Update / Create XB's if the content contains XB's.
+	if ( ! empty( $xbs ) ) {
+		update_create_xb( $xbs, 'Widget', null, $widget->number, 'widget' );
```
roborourke, line 123: User facing text should be translatable:
```suggestion
		update_create_xb( $xbs, _( 'Widget' ), null, $widget->number, 'widget' );
```
I think core WP probably has a translation for "Widget" on its own so we can omit the text domain argument.

inc/blocks/namespace.php
```diff
@@ -93,6 +95,47 @@ function on_save_post( int $post_ID, WP_Post $post, bool $update ) : void {
...
+/**
+ * Update or create new XB's
+ *
+ * @param array $xbs XB's on page or as widgets.
+ * @param string $default_title Default Title title for XB Post.
+ * @param int $post_ID Post ID if applicable in use case.
+ * @param int $item_id Item ID.
+ * @param string $context Meta Key for updating the post_meta.
+ */
+function update_create_xb( array $xbs, string $default_title = '', ?int $post_ID = null, ?int $item_id = null, string $context = 'post' ) {
```
roborourke, line 138: Minor thing but for function names it's fine to be a bit more wordy if it makes the purpose more obvious e.g. `update_or_create_xb_post()`.

This is sometimes known as an "upsert" too e.g. update if thing exists else insert thing, but that terminology isn't used anywhere in WP.

inc/blocks/namespace.php
```diff
@@ -130,16 +173,21 @@ function on_save_post( int $post_ID, WP_Post $post, bool $update ) : void {
 				'post_content' => serialize_block( $xb ),
 				'post_name' => $xb['attrs']['clientId'],
 				'post_author' => get_current_user_id(),
-				'post_title' => $xb['attrs']['title'] ?: $default_title,
+				'post_title' => $xb['attrs']['title'] ?: $xb_title,
 				'post_parent' => $post_ID,
+				'meta_input' => [
```
roborourke, line 178: I didn't know this was a thing to be honest! Might as well move the `_xb_type` meta values here too.

roborourke (commented):
Great work! Couple of minor bits of polish and this is good to go

src/ab-test-preview.js
robindevitt: I had done this slightly differently further done in the function, but this method makes more sense.

inc/blocks/namespace.php
robindevitt: I have now added the `_xb_type` meta values here too.

inc/blocks/namespace.php
```diff
@@ -93,6 +95,47 @@ function on_save_post( int $post_ID, WP_Post $post, bool $update ) : void {
...
+/**
+ * Update or create new XB shadow posts.
+ *
+ * @param array $xbs XBs on page or as widgets.
+ * @param string $default_title Default Title title for XB Post.
+ * @param int $post_ID Post ID if applicable in use case.
+ * @param int $item_id Item ID if applicable in use case.
+ * @param string $context Meta Key for updating the post_meta.
+ */
+function update_or_create_xb_post( array $xbs, string $default_title = '', ?int $post_ID = null, ?int $item_id = null, string $context = 'post' ) {
```
robindevitt, line 138: @roborourke This Im not sure if you maybe have a suggestion to improve it better. 

The `$post_ID` variable which is used for the `post_parent` when doing the `wp_insert_post` and `wp_update_post`.

Then the `item_id` variable is used for when the XBs are being used in widgets. It's then added to the `$xb_title` value to differentiate the different titles.

Not sure if I am being a bit to pedantic on this change?

roborourke: I think this makes sense, it's not going to get loads of additional use and it's internal so it can always be iterated on but this approach is specific about what the arguments are in a good way without overloading anything too much. I like it.

inc/blocks/ab-test/edit.js
```diff
@@ -192,7 +192,7 @@ const Edit = ( {
 					{ currentPost.type !== 'xb' && (
 						<TextControl
 							label={ <strong>{ __( 'Block Title', 'altis-analytics' ) }</strong> }
-							placeholder={ sprintf( __( '%s (XB %d)', 'altis-analytics' ), currentPost.title, instance.current ) }
+							placeholder={ sprintf( '%s %d', ( currentPost.title || __( 'A/B Test', 'altis-analytics' )  ), instance.current ) }
```
roborourke, line 195:
```suggestion
							placeholder={ sprintf( '%s %s %d', decodeEntities( currentPost.title || '' ), __( 'A/B Test', 'altis-analytics' ), instance.current ).trim() }
```

roborourke (approved):
Couple of minor thoughts I had and a modification to the `_xb_type_<name>` meta key, nice one @robindevitt, this was a tricky one.

PR #315, opened 2022-01-11
Fix A/B tests defaults tests

shadyvb:
This fixes an issue where a default object in JS was not being utilized as expected, so the defaults set there was not being respected by the front-end, rather the ones comes from server-side/API.

Removed the defaults in JS, apart from one location where the test needs to be reset on the front-end.

### DIFF
--- a/inc/experiments/featured-images/namespace.php
+++ b/inc/experiments/featured-images/namespace.php
@@ -27,7 +27,6 @@ function setup() {
 function init() {
 	if ( ! is_admin() ) {
 		add_filter( 'post_thumbnail_html', __NAMESPACE__ . '\\filter_post_thumbnail_html_ab_test_values', 10, 5 );
-		add_filter( 'has_post_thumbnail', __NAMESPACE__ . '\\filter_has_post_thumbnail', 10, 2 );
 	}
 
 	Experiments\register_post_ab_test(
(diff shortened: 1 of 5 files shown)


inc/experiments/featured-images/namespace.php
```diff
@@ -10,6 +10,8 @@
 use Altis\Analytics\Experiments;
 use Altis\Analytics\Utils;
 
+use function Altis\Analytics\Experiments\is_ab_test_running_for_post;
+
```
roborourke, line 14:
```suggestion
```

inc/experiments/namespace.php
```diff
@@ -655,7 +655,8 @@ function is_ab_test_started_for_post( string $test_id, int $post_id ) : bool {
  * @return int A percentage
  */
 function get_ab_test_traffic_percentage_for_post( string $test_id, int $post_id ) : int {
-	return (int) get_post_meta( $post_id, '_altis_ab_test_' . $test_id . '_traffic_percentage', true );
+	$traffic_percentage = get_post_meta( $post_id, '_altis_ab_test_' . $test_id . '_traffic_percentage', true );
+	return (int) is_numeric( $traffic_percentage ) ? $traffic_percentage : 35;
```
roborourke, line 659: Does this cast the return value of `is_numeric()` to an int or the return value of the ternary? Maybe to be safe and explicit it can be:
```suggestion
	return (int) ( is_numeric( $traffic_percentage ) ? $traffic_percentage : 35 );
```

roborourke (approved):
Two suggestions on the traffic percentage return value and an unused use statement to resolve.

Otherwise if you're happy that there are no errors or anything when loading a new post edit screen with no existing data then this is good to go.

PR #325, opened 2022-01-19
Avoid fatal error with collect_aggs returning null

roborourke:
In some cases `collect_aggregations()` returned `null`, and when that was passed to `relativize_urls()` the type check would cause a fatal error.

### DIFF
--- a/inc/dashboard/namespace.php
+++ b/inc/dashboard/namespace.php
@@ -627,11 +627,11 @@ function register_default_aggregations() {
  * Process a terms aggregation into key value pairs.
  *
  * @param array|null $aggregation A terms aggregation result from Elasticsearch.
- * @return array|null
+ * @return array
  */
-function collect_aggregation( ?array $aggregation ) : ?array {
+function collect_aggregation( ?array $aggregation ) : array {
 	if ( empty( $aggregation ) ) {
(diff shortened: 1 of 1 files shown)


roborourke:
Linter errors are unrelated to this change but we should fix those. New CSS introduced recently.

inc/dashboard/namespace.php
```diff
@@ -1169,7 +1169,7 @@ function get_top_data( $start, $end, ?Filter $filter = null ) {
 				'name' => get_the_author_meta( 'display_name', $post->post_author ),
 				'avatar' => get_avatar_url( $post->post_author ),
 			],
-			'views' => $processed[ $post->ID ]['total'],
+			'views' => $processed[ $post->ID ]['total'] ?? 0,
```
mikelittle, line 1172: Is this an unrelated just-in-passing fix?

mikelittle (approved):
One question but not an issue. Looks good to me

inc/dashboard/namespace.php
roborourke: Forgot to do `git commit -p` 🤦‍♂️  - It is yeah. May as well leave it here now.

Issue #317, opened 2022-01-11
Data export can show invalid JSON for an empty response with response code 200


roborourke:
Sometimes the data export endpoint returns no results but JSON output like the following with a 200 response code:

```
[
    ,
]
```

Ways to reproduce:

- Run the export with Elasticsearch stopped
- Run the export with no entries in elasticsearch (use a DELETE request to `/analytics-*` via Kibana dev tools
- Run the export for a day in the future or a day with no data in ES

shadyvb:
@roborourke can we add some reproduction steps here ?

roborourke:
Added some ways to reproduce it in the original comment.

robindevitt:
Hey @roborourke 

> Run the export with Elasticsearch stopped

How does one stop Elasticsearch?

> Run the export with no entries in elasticsearch (use a DELETE request to /analytics-* via Kibana dev tools

Is this the correct documentation for [Kibana dev tools](https://www.elastic.co/guide/en/kibana/current/devtools-kibana.html)? I came across [this](https://docs.altis-dxp.com/local-server/elasticsearch/#kibana) in the Altis Docs.

Thanks
Robin

roborourke:
I think the images on the Altis docs are out of date but yes. There’s a link to Kibana in the Tools menu in the admin you can click, then look for Dev Tools in the left hand menu

robindevitt:
This has been completed and merged: [Issue](https://github.com/humanmade/aws-analytics/issues/317) & [PR](https://github.com/humanmade/aws-analytics/pull/333)

PR #333, opened 2022-02-09
Add a check to ensure event source data is not empty.

robindevitt:
### Issue #317 
Sometimes the data export endpoint returns no results but JSON output like the following with a 200 response code:
```
[
    ,
]
```

### findings & code reference
The total pages is always set to a minimum of 2 as Elasticsearch slices need at least 2 pages and no more than 1024 by default.
https://github.com/humanmade/aws-analytics/blob/d7f5e4029bdca8afad74eca60bcf3c6390021d3b/inc/export/class-endpoint.php#L145-L147

It appeared that there was a comma ( , ) being added between result sets except for the last page.
https://github.com/humanmade/aws-analytics/blob/d7f5e4029bdca8afad74eca60bcf3c6390021d3b/inc/export/class-endpoint.php#L226-L227

### Solution
When creating the output a comma will only be aded in between result sets when there is event source data. The comma will continue to not be added to the last page.

### DIFF
--- a/inc/export/class-endpoint.php
+++ b/inc/export/class-endpoint.php
@@ -220,11 +220,13 @@ class Endpoint {
 			// Extract the event source data only.
 			$events = wp_list_pluck( $results['hits']['hits'], '_source' );
 
-			if ( ! empty( $events ) && $format === 'json' ) {
+			if ( $format === 'json' ) {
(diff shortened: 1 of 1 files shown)


roborourke (approved):
Tidy, cheers @robindevitt!

PR #342, opened 2022-02-18
Add polyfill.io bundle locally

shadyvb:
Ref https://github.com/humanmade/product-dev/issues/886


roborourke (approved):

rmccue:
@roborourke @shadyvb Looks like this never got backported to the active branch?

shadyvb:
I had no idea how branching works here. Good discussion to be had.

PR #345, opened 2022-03-04
Standalone filters

joehoyle:
Support a couple filters etc to run aws-analytics standalone/

### DIFF
--- a/inc/utils/namespace.php
+++ b/inc/utils/namespace.php
@@ -193,9 +193,14 @@ function get_elasticsearch_version() : ?string {
 	return $version;
 }
 
-function get_elasticsearch_index_path() : string {
-	$index_path = apply_filters( 'altis.analytics.elasticsearch.index-path', 'analytics-' );
-	return $index_path;
+/**
+ * Get the index name prefix for analytics indexes.
+ *
+ * @return string
+ */
+function get_elasticsearch_index_prefix() : string {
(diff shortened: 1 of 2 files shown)


inc/utils/namespace.php
```diff
@@ -193,6 +193,11 @@ function get_elasticsearch_version() : ?string {
 	return $version;
 }
 
+function get_elasticsearch_index_path() : string {
```
rmccue, line 196: Needs phpdoc

inc/utils/namespace.php
```diff
@@ -193,6 +193,11 @@ function get_elasticsearch_version() : ?string {
...
+	$index_path = apply_filters( 'altis.analytics.elasticsearch.index-path', 'analytics-' );
```
rmccue, line 197: `prefix` seems like it'd be better?

inc/utils/namespace.php
```diff
@@ -204,7 +209,8 @@ function get_indices() : array {
 		return $cache;
 	}
 
-	$indices_response = wp_remote_get( get_elasticsearch_url() . '/analytics-*?filter_path=*.aliases' );
+	$index_path = apply_filters( 'altis.analytics.elasticsearch.index-path', 'analytics-' );
```
rmccue, line 212: As above

rmccue (changes requested):

shadyvb (approved):

PR #349, opened 2022-03-14
Don't load experiments sidebar if no tests for current post type

robindevitt:
### Issue
The [original issue](https://github.com/humanmade/altis-cms/issues/536) highlights steps to replicate the reported issue.

### Initial findings
Currently it appears that the `Featured image` and `Title` experiments are setup with the following which explains why the error is occurring when adding a post of another `post_type`, namely the `wp_block` which is the post type of Reusable Blocks.
```
'post_types' => [
	'post',
	'page',
],
``` 

### Solution/s
1. Adding support for the post type:
```
'post_types' => [
	'post',
	'page',
	'wp_block',
],
```

2. Having a function that determines the post type and which the experiment should be added or not. This could help with other post types available or future post types that may be supported.

Currently I have only pushed a solution for point 1.

### DIFF
--- a/inc/experiments/namespace.php
+++ b/inc/experiments/namespace.php
@@ -472,7 +472,7 @@ function enqueue_experiments_editor_scripts( string $hook ) : void {
 	);
 
 	// Retrieve the labels for tests that show UI.
-	$js_data = array();
+	$js_data = [];
 
 	foreach ( $post_ab_tests as $test_id => $test ) {
 		if ( empty( $test['editor_scripts'] ) ) {
--- a/inc/experiments/namespace.php
+++ b/inc/experiments/namespace.php
@@ -1261,27 +1261,3 @@ function end_ab_test_for_post( string $test_id, int $post_id ) {
 	 */
 	do_action( "altis.experiments.test.ended.{$test_id}", $post_id );
 }
-
-/**
- * Check if the post type is supported.
- *
- * @param array $supported_post_types The supported post types of the experiment.
- * @return bool True if the post type is not supported.
- */
-function post_type_support( array $supported_post_types ) {
-	global $pagenow;
-	$post_type = '';
-
-	// Set the variable $post_type from the post type in the page url.
-	if ( isset( $_GET[ 'post_type' ] ) && !empty( isset( $_GET[ 'post_type' ] ) ) ) {
-		$post_type = $_GET[ 'post_type' ];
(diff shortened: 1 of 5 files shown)


robindevitt:
@shadyvb Please can you have a look here at my solution/s and let me know if there could be a better solution of if my suggested solution in point #1 will do. Thank you!

robindevitt:
I came across [this function](https://github.com/humanmade/aws-analytics/blob/0dbf13532682c3555fa5eee529331d6626cf78a4/inc/experiments/namespace.php#L326-L413)

Very specifically this section which loops through the `post_types` in the variable `$options` and then adds post type support for each of the post types listed. This explains why my solution point 1 works.
```
// Handle post type support filtering and registration.
foreach ( $options['post_types'] as $post_type ) {
	add_post_type_support( $post_type, "altis.experiments.{ $test_id }" );
}
``` 

@shadyvb if I understand your comments correctly, what needs to happen is if the experiments don't support the post type then the experiments should be shown?

shadyvb:
Yup, that seems sensible.

robindevitt:
@shadyvb I have pushed up further changes for the experiments and post type support. Let me know if this approach is better suited. Thank you 🙌🏼

shadyvb:
@robindevitt were you able to test your changes ? eg: test that the bug is fixed AND that the feature still works ?

inc/experiments/namespace.php
```diff
@@ -483,13 +486,15 @@ function enqueue_experiments_editor_scripts( string $hook ) : void {
...
 		}
-	}
 
-	$js_data = array_filter( array_map( function( array $test ) {
-		return $test['show_ui']
-			? array_intersect_key( $test, array_flip( [ 'label', 'singular_label' ] ) )
-			: false;
-	}, $post_ab_tests ) );
+		if ( $test['show_ui'] ) {
+			$js_data[$test_id] = [
```
roborourke, line 491: Minor issue on the formatting, we use spaces around array keys if it's a variable:
```suggestion
			$js_data[ $test_id ] = [
```

roborourke (approved):
Perfect, I'll commit the code style suggestion but this is the right solution 👍

inc/experiments/namespace.php
robindevitt: Thanks for catching that @roborourke , I had hoped to run the HM Linter before submitting the code review but it wasn't showing.

PR #372, opened 2022-04-07
Altis dashboard pagination wrapping.

robindevitt:
#### Issue 
[Original issue](https://github.com/humanmade/altis-analytics/issues/239)
The pagination links are not wrapped and go outside the visible screen when there are a large number of pages.
157015083-04873af5-6021-4550-bd89-6da30666fa50

#### Solution
The solution was provided in the initial issue to adding `flex-wrap:wrap;` to the container to resolve the issue.
Screenshot 2022-04-07 at 16 18 54


roborourke:
I think we may need to add some logic to limit the number of visible pages too, I'll write up some acceptance criteria

robindevitt:
In a use case where a user has 200+ pages on their site. Pagination will start "1 2 3 4 ... 398 399 400" for example. If the user was needing to navigate to page 150, they would need to click through roughly 147 times until they can click page 150, would having an option for a user to select a page or input a page number to navigate to quickly be useful too?

rmccue:
I don't think we necessarily need a jump-to-page capability here; with that much data, search is more useful than jumping to a page. Plus, if you're sorting by popularity, you wouldn't know what page it's on.

If we want to keep it anyway, we can combine with the page indicator, eg "Page \_2\_ of 12 [Go]"

robindevitt:
I have dropped the option to navigate to a specific page due to the limited options available. 

This is what it currently look like:

robindevitt:
@roborourke This is ready for review, I had you as a reviewer already but couldn't re-request a review.

src/accelerate/components/List.tsx
```diff
@@ -1,6 +1,7 @@
 import React, { useState } from 'react';
 import { useSelect } from '@wordpress/data';
 import { __ } from '@wordpress/i18n';
+import { Pagination } from "react-pagination-bar"
```
roborourke, line 4:
```suggestion
import { Pagination } from 'react-pagination-bar';
```
Single quotes for imports and always use a semi-colon. If in doubt look at the existing code style in the file and keep it consistent with that.

src/accelerate/components/List.tsx
```diff
@@ -39,6 +40,9 @@ export default function List( props: Props ) {
 
 	const maxViews = posts.reduce( ( carry, post ) => Math.max( post.views, carry ), 0 );
 
+	const [currentPage, setCurrentPage] = useState(1);
```
roborourke, line 43: Spacing
```suggestion
	const [ currentPage, setCurrentPage ] = useState( 1 );
```

src/accelerate/components/List.tsx
```diff
@@ -39,6 +40,9 @@ export default function List( props: Props ) {
...
+  	const pagePostsLimit = 3;
```
roborourke, line 44: This doesn't appear to be used anywhere

src/accelerate/components/List.tsx
```diff
@@ -156,33 +160,22 @@ export default function List( props: Props ) {
...
-							} ) }
-							<button
-								disabled={ page <= pagination.pages }
-								type="button"
-								onClick={ () => setPage( page + 1 ) }
-							>
-								{ __( 'Next', 'altis' ) } →
-							</button>
+							{ pagination.total > 0 && (
+								<>
```
roborourke, line 164: Given that if this doesn't show then we'll have empty `.table-footer` and `.pagination` divs. We could move this logic higher so the pagination div at least is only output if we have a non zero `pagination.total` value.

You wouldn't need the fragment wrapper then.

src/accelerate/components/List.tsx
```diff
@@ -156,33 +160,22 @@ export default function List( props: Props ) {
...
+										itemsPerPage={25}
+										onPageСhange={ ( pageNumber ) =>{
+											setCurrentPage( pageNumber );
+											setPage( pageNumber );
+										} }
+										totalItems={pagination.total}
+										pageNeighbours={5}
+										withGoToInput={true}
+									/>
+									<span className="current-page">{ __( 'Page ', 'altis' ) } { page } { __( 'of ', 'altis' ) } { pagination.pages }</span>
```
roborourke, line 176: Check you spacing and formatting all throughout here. The dev server should be warning you about some of it but if not it's the spacing of prop values and removing the superfluous brackets around the single parameter in `onPageChange` arrow function that needs updating.

roborourke (changes requested):
Looking alright but some bits that could use a bit of polish

src/accelerate/components/List.tsx
robindevitt: Oversight from my side, when I was implementing the new pagination I kept the old pagination in place, removed and resolved now.

roborourke (approved):
One small edit on the page change callback

PR #383, opened 2022-05-02
Mis-styled text on insights page/popup

robindevitt:
This is a solution using CSS with minimal changes.

### Issue
When accessing the insights view for an A/B block there is a mis-styled date.

[Original issue ](https://github.com/humanmade/product-dev/issues/998)

As a popup from within the post editor it looks like this:
Screenshot 2022-05-02 at 17 39 11

### Testing notes
I used a fixed width, which was determined using using a "potential" longest date of "September 33, 3333" to ensure the date stays on a single line.

The negative margins allow the span in the original issue to remain centered as the slider moves during the timeline.  

### Solution
Suggested changes will allow it to look like this:
Screenshot 2022-05-02 at 17 41 36
Screenshot 2022-05-02 at 17 42 16

### DIFF
--- a/src/blocks/ui/components/timeline.js
+++ b/src/blocks/ui/components/timeline.js
@@ -64,6 +64,7 @@ const StyledTimeline = styled.div`
 			text-align: right;
 			bottom: -4px;
 			padding-bottom: 4px;
+			white-space: nowrap;
 			&:before {
 				top: auto;
 				bottom: -7px;
@@ -72,9 +73,6 @@ const StyledTimeline = styled.div`
 
 			.altis-analytics-timeline__text {
 				margin-right: -2.5rem;
-				margin-left: -2.5rem;
-				width: 130px;
-				display: inline-block;
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


robindevitt:
I have done the CSS changes inline via my browser, however I have an "issue" getting it to show on my local.

When clicking the "View experience insights" button I have the following error show-
> This block has encountered an error and cannot be previewed.

roborourke:
I don’t think this is really what we’re after, it looks like the test is already part way through. Is there a way to set a minimum width on the container for the current date? Or use white-space: no wrap on it?

robindevitt:
Thanks for the feedback @roborourke.

I have removed my initial changes as it was achievable with `white-space: no wrap;`.

roborourke (approved):

src/blocks/ui/components/timeline.js
```diff
@@ -64,6 +64,7 @@ const StyledTimeline = styled.div`
 			text-align: right;
 			bottom: -4px;
 			padding-bottom: 4px;
+			white-space: nowrap;
```
mikelittle, line 67: Do! Is that all it was? I thought I'd tried that, but maybe I tried it in the wrong place! Cheers.

PR #384, opened 2022-05-03
Prevent fatal error if aggregations are empty

roborourke:
Noticed when there's no analytics data or indexes yet, rare but can happen at least on local.


robindevitt (approved):
Looks good 👍🏼

PR #389, opened 2022-05-06
Switch to intersection observers for tracking XB views

roborourke:
There was a problem with the earlier scroll tracking behaviour not triggering on first load, so some events were not being picked up.

Additionally once that was resaolved I observed double tracking of events caused by not clearing the queue early enough and a race-condition happening.Needs further work in a future update to make delivery more robust.

Fixes https://github.com/humanmade/product-dev/issues/1004

### DIFF
--- a/src/analytics.js
+++ b/src/analytics.js
@@ -746,6 +746,10 @@ const Analytics = {
 	 * @param {object} endpoint Optional updated endpoint data.
 	 */
 	flushEvents: async ( endpoint = {} ) => {
+		// Snapshot events to send and clear.
+		const eventsToDeliver = Analytics.events;
+		Analytics.events = [];
+
 		// Ensure flushEvents isn't called too quickly when set via timeout.
 		if ( Analytics.timer ) {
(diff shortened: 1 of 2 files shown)


roborourke:
If you're happy with the changes @mikelittle go ahead and merge, then merge the following backport PRs if they are created successfully. I'll go through and correct any that don't work automatically.

Once you have merged the backported PRs and you feel comfortable doing so you can go to the releases page on this repo and create a new release from each version branch you merged a backport to.

For example select `v4-6-branch` and enter a new tag name with the patch version bumped e.g. 4.6.2 and choose "create new tag on publish".

Enter the tag version number as the release title and then this PR title as the description entry then hit "publish release". The assets will be built automatically.

No worries if you prefer not to, happy to go through the process with you next week.

shadyvb (approved):
LGTM, one note though on use of `IntersectionObserver` is that it breaks compatibility with IE, so we might need to factor that in the release version / docs.

roborourke:
Added the polyfill.io code for intersection observer, good catch 👍

PR #408, opened 2022-05-18
Handle cron schedules being null

joehoyle:
We had a case where some 3rd party code is setting `$schedules` to null.

### DIFF
--- a/inc/export/cron/namespace.php
+++ b/inc/export/cron/namespace.php
@@ -73,7 +73,10 @@ function bootstrap() : void {
  * @return array
  */
 function setup_cron_interval( $schedules ) : array {
-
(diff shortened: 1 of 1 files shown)


tfrommen (approved):

roborourke:
Does this affect more than just 4.6? We may need to backport it further

joehoyle:
IMO no, it was a fix for a bad-acting filter adjusting the type in accelerate. I don't think it's worth backporting-- though might be one data point to evaluate whether type hinting should be used for filter / action callbacks at all.

roborourke:
Interesting. The alternative is type checking in every filter I suppose, but then do you trust the value or return a default... It can be useful feedback during development to get a fatal. Tricky one to apply a blanket approach to.

PR #410, opened 2022-05-19
Don't show metrics when the value is "infinity".

robindevitt:
[Original Issue ](https://github.com/humanmade/product-dev/issues/1020)

#### issue 
If comparing against a previous value of 0, the dashboard insights screen shows an increase of "infinity percent".

Screenshot 2022-05-19 at 08 09 17

#### Solution
In the original issue it was determined to display nothing in these cases. This PR's solution adds a check that determines the if the prop received is an integer. If it is an integer it displays the metrics.

Screenshot 2022-05-19 at 08 34 53

### DIFF
--- a/src/accelerate/components/Insight.tsx
+++ b/src/accelerate/components/Insight.tsx
@@ -20,7 +20,7 @@ export default function Insight( props: Props ) {
 				</div>
 				<div className="key-insight-metrics">
 					<div className="metrics-aggregate">{ compactMetric( props.stat || 0 ) }</div>
-					{ !! props.delta && ! isNaN( props.delta ) && Number.isInteger( props.delta ) && (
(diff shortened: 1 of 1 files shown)


mikelittle (commented):
Is `isInteger` the correct test here? I see the same property is later parsed as a float.  
It looks like if the value was 10.5 is would not display?

robindevitt:
@mikelittle , thank you for the suggest and direction. I have gone with using `isFinite` rather than `isInteger`.

roborourke:
Is that the only place this output can appear?

robindevitt:
I haven't seen it anywhere else other than on the dashboard metric for `Page Views`, `Experienced Block views` and `Lift`.

PR #415, opened 2022-06-17
Analytics module events to track in Altis

robindevitt:
Related issue https://github.com/humanmade/product-dev/issues/996

### DIFF
--- a/src/audiences/ui/edit.js
+++ b/src/audiences/ui/edit.js
@@ -107,8 +107,22 @@ class Edit extends Component {
 		// Update if we have an ID, otherwise create a new one.
 		if ( post.id ) {
 			onUpdatePost( post );
+			analytics.track(
+				'update',
+				{
+					content_type: 'audience',
+					status: post.status,
+				}
+			);
 		} else {
 			onCreatePost( post );
+			analytics.track(
+				'create',
+				{
+					content_type: 'audience',
+					status: post.status,
+				}
+			);
 		}
 
 		// Edits have been saved now so it's safe to navigate again.
--- a/src/audiences/ui/manager.js
+++ b/src/audiences/ui/manager.js
@@ -62,6 +62,12 @@ class Manager extends Component {
 					onClick: () => {
 						onSetCurrentPost( defaultPost );
 						this.setState( { view: 'edit' } );
+						analytics.track(
+							'initiate_create',
+							{
+								content_type: 'audience',
+							}
+						);
 					},
 				},
 				/**


robindevitt:
I managed to get this hooked up to Segment with a free account to do some initial tests. So far everything looks pretty good.

I noticed a minor discrepancy in some places the date ranges have different "values". Two examples below:

```javascript
analytics.track('filter', {
  filter_type: 'date range',
  filter_value: '90 Days',
  location: 'dashboard'
});
```
& 
```javascript
analytics.track('filter', {
  filter_type: 'date range',
  filter_value: 90,
  location: 'insights'
});
```

Should I work on that when ever it's a date range it will always be the `number` and include "days" ?

robindevitt:
Hey @roborourke 

Thanks for the suggestions and direction. I have pushed changes now.

```javascript
analytics.track(
	__( 'filter', 'altis-analytics' ),
	{
		location: __( 'dashboard', 'altis-analytics' ),
		filter_type: __( 'author', 'altis-analytics' ),
		filter_value: e.target.options[e.target.options.selectedIndex].text,
	}
);
```
In the above example the select hasn't got a name attribute. So would the structure and names/labels be appropriate?

roborourke:
> In the above example the select hasn't got a name attribute. So would the structure and names/labels be appropriate?

Are you referring to the code example? Then yeah, although the values should not be translatable. That would make querying in Mixpanel near impossible!

src/accelerate/components/List.tsx
```diff
@@ -73,6 +81,14 @@ export default function List( props: Props ) {
 							onChange={ e => {
 								setPage( 1 );
 								setUser( parseInt( e.target.value, 10 ) );
+								analytics.track(
+									__( 'filter', 'altis-analytics' ),
+									{
+										location: __( 'dashboard', 'altis-analytics' ),
+										filter_type: __( 'author', 'altis-analytics' ),
+										filter_value: e.target.options[e.target.options.selectedIndex].text,
```
roborourke, line 89: Again, is this the right value to be using here?

src/blocks/ui/components/date-range.js
```diff
@@ -28,27 +28,36 @@ const Form = styled.form`
...
+						value={ range.value }
 						onChange={ e => {
-							e.target.checked && onSetRange( range );
+							e.target.checked && onSetRange( range.value );
+							analytics.track(
+								__( 'filter', 'altis-analytics' ),
+								{
+									location: __( location, 'altis-analytics' ),
+									filter_type: __( 'date_range', 'altis-analytics' ),
+									filter_value: range.label.replace( / /g, '_' ),
```
roborourke, line 54: Again, why is this the label?

src/blocks/ui/insights/personalization.js
```diff
@@ -37,6 +37,21 @@ const Personalization = ( {
 		};
 	}, [ clientId ] );
 
+	const ranges = [
```
roborourke, line 40: This seems to more or less duplicate the periods object we already have, in addition the labels here are not translatable. It would be better to reuse the periods object, and make the `DateRange` component work with the same values e.g. `P7D`, `P30D` and `P90D`

src/blocks/ui/components/date-range.js
```diff
@@ -28,27 +28,36 @@ const Form = styled.form`
...
+								}
+							);
 						} }
 					/>
-					<label htmlFor={ `altis-analytics-date-range-${ range }` }>
-						{ sprintf( __( '%d days', 'altis-analytics' ), range ) }
+					<label htmlFor={ `altis-analytics-date-range-${ range.value }` }>
+						{ __( range.label, 'altis-analytics' ) }
```
roborourke, line 60: Variables cannot be used with translation functions.

src/blocks/ui/insights/personalization.js
```diff
@@ -47,7 +62,8 @@ const Personalization = ( {
 	return (
 		<>
 			<div className="altis-analytics-block-metrics">
-				<DateRange ranges={ [ 7, 30, 90 ] } value={ days } onSetRange={ setDays } />
+				<DateRange
+					location="insights" ranges={ ranges } value={ days } onSetRange={ setDays } />
```
roborourke, line 66: If your editor is et to have a max line limit it's probably better to put all the props on new lines rather than split it like this.

```suggestion
				<DateRange
					location="insights"
					ranges={ ranges }
					value={ days }
					onSetRange={ setDays }
				/>
```

src/accelerate/components/Hero.tsx
```diff
@@ -44,10 +52,38 @@ export default function Hero( props: Props ) {
...
+									onClick={ () => {
+										analytics.track(
+											__( 'click', 'altis-analytics' ),
+											{
+												location: __( 'dashboard', 'altis-analytics' ),
+												url: 'edit.php?post_type=xb',
+												link_text: __( 'Insights', 'altis-analytics' ),
+											}
+										);
+									}}
```
roborourke, line 83:
```suggestion
									} }
```

src/accelerate/components/Hero.tsx
```diff
@@ -44,10 +52,38 @@ export default function Hero( props: Props ) {
 					{ !! props.period && ( props.canViewAnalytics || props.canViewInsights ) && (
 						<nav className='Hero__links'>
 							{ props.canViewAnalytics && (
-								<a href="index.php?page=altis-analytics">{ __( 'Analytics', 'altis-analytics' ) }</a>
+								<a
```
roborourke, line 55: One option could be to make a new component for links that has the tracking built in, so some props could be provided only once e.g. `url` or `href` etc... Not required but always something to consider when you spot some repetition in the code.

roborourke (changes requested):
Some overall stuff to get started on here, primarily around the misuse of translation functions, and double checking what we're tracking.

There are some opportunities for code reuse as well worth exploring.

robindevitt:
Thanks @roborourke, would it be better if I didn't translate all the items sent to Segment? Or should I use it more sparingly?

roborourke:
> Thanks @roborourke, would it be better if I didn't translate all the items sent to Segment? Or should I use it more sparingly?

For sending data to segment we shouldn’t use translation at all. We could end up with different values for the same thing

inc/audiences/namespace.php
```diff
@@ -249,6 +251,49 @@ function save_post( $post_id ) {
 
 	// phpcs:ignore HM.Security.ValidatedSanitizedInput.InputNotSanitized -- Sanitized in save_audience().
 	save_audience( $post_id, wp_unslash( $_POST['audience'] ) );
+
+	// Track post status creation
+	do_action( 'altis.telemetry.track', [
+		'event' => __( 'create', 'altis-analytics' ),
```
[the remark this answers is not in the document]
robindevitt: Thank you for catching this! I misread the docs on `save_post`, I have corrected it now.

What I have picked up on too is the following in Altis Core where we are currently tracking when posts are saved / updated. We will essentially be tracking audiences updates & creation twice :

https://github.com/humanmade/altis-core/blob/6adb23f1c65c5ed33622c52670682ca3a6f1064d/inc/telemetry/namespace.php#L500-L527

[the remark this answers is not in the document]
roborourke: Yeah I was thinking we could update the way it's done in core to be more generic and handle it there rather than in this plugin.

src/blocks/ui/insights/personalization.js
robindevitt: @roborourke I notice the periods object is located within a [`.ts` file](https://github.com/humanmade/aws-analytics/blob/master/src/data/periods.ts). I suspect this can't be imported to a [`.js` file](https://github.com/humanmade/aws-analytics/blob/master/src/blocks/ui/insights/personalization.js) and I would need to recreate a new object is that correct?

src/accelerate/components/Hero.tsx
robindevitt: Good thinking and totally agreed ✅

inc/audiences/namespace.php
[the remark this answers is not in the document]
roborourke: Lets discuss this in slack tomorrow - remind me at standup!

src/blocks/ui/insights/personalization.js
roborourke: Ah right yes, we have the split build processes still... my usual approach would just be to try it out and see but I think you're right.

In which case it might be good to make a non typescript copy of the periods object and use that, with a comment noting that it should use the typescript version in future.

robindevitt: @roborourke I got this setup and running but before I push my current changes up, given that the `DateRange`  component is only being used once and would essentially be replaced at later stage, would it not be better to keep the duplicate and get it to work rather than duplicating the typescript copy?

src/accelerate/components/List.tsx
robindevitt: So currently the filter only allows the select of `All Authors` and `My Content`. The respect values are ` ` and `1` which I figured would have no real value so I thought it was best to use the option select. Potentially a better solution to use?

src/data/periods.js
```diff
@@ -0,0 +1,26 @@
...
+		number: 30,
+	},
+	{
+		label: __( '90 days', 'altis' ),
+		value: 'P90D',
+		diff: null,
+		number: 90,
+	},
+];
+export default Periods;
```
robindevitt, line 26: 1. I created this for the periods to work with the date range. I added `number` as the number value is needed for the request for the posts. It wasn't accepting `P7D` as a value only `7`. 

2. What is the purpose of the `diff` value as I wasn't able to identify where/what it's used for?

robindevitt:
@roborourke When you get a gap, I have gone through your previous review and pushed up the relevant changes. 

I have added a comment on the changes for the `DateRange` with 2 questions around the structure. 

Lastly [the values used for the `Author` filtering](https://github.com/humanmade/aws-analytics/pull/415%23discussion_r903659228) , I'm not sure what/if we want to action anything further?

src/accelerate/components/Link.tsx
```diff
@@ -0,0 +1,28 @@
...
+					'click',
+					{
+						location: props.location,
+						url: props.url,
+						link_text: props.linkText,
+					}
+				);
+			}}
+		>
+			{ __( props.linkText, 'altis-analytics' ) }
```
roborourke, line 25: Variables can't be used in translation functions - they are analysed / extracted from code using static analysis.

src/accelerate/components/List.tsx
roborourke: This should be  using `e.target.value` as the `filter_value` value

src/accelerate/components/List.tsx
```diff
@@ -94,6 +111,15 @@ export default function List( props: Props ) {
...
 									setSearch( value );
 								}, 500, e.target.value );
+								analytics.track(
+									'filter',
+									{
+										location: 'dashboard',
+										filter_type: 'search',
+										filter_value: e.target.value,
+									}
+								);
```
roborourke, line 121: Lets put this inside the `setTimeout` callback or we're going to be tracking events on every keystroke

src/blocks/ui/components/date-range.js
roborourke: This still needs fixing

src/data/periods.js
```diff
@@ -0,0 +1,26 @@
+import { __ } from '@wordpress/i18n';
+
+/**
+ * Date Periods. This is here to accomodate the split build process. The typescript copy should supersede this when appropriate.
+ */
+const Periods = [
```
roborourke, line 6: Typically the capitalisation is reserved for react components and names of type definitions. I think it may have been confusing where you saw `const periods: Periods = [ .... ]` - `periods` is the variable name, and `Periods` is the type definition name.

src/data/periods.js
roborourke: 1. I was suggesting updating the API to accept values like `P7D` rather than just numbers so we were able to do this and standardise on one approach
2. `diff` is used to indicate the time range to get data for to calculate the percentage change comparison, e.g. when showing data for the last 7 days, compare it to the 7 days before that.

roborourke (changes requested):
Still some edits to make here where we still have variables in translation functions for example.

src/accelerate/components/List.tsx
```diff
@@ -110,16 +110,15 @@ export default function List( props: Props ) {
 								timer && clearTimeout( timer );
 								timer = setTimeout( value => {
 									setSearch( value );
+									analytics.track(
+										'filter',
+										{
+											location: 'dashboard',
+											filter_type: 'search',
+											filter_value: e.target.value,
```
roborourke, line 118: Weirdly `e.target.value` might actually be different than what is intended here, this is why you can pass values to the `setTimeout` callback, use `value` instead:
```suggestion
											filter_value: value,
```

src/accelerate/components/Hero.tsx
```diff
@@ -44,10 +53,18 @@ export default function Hero( props: Props ) {
 					{ !! props.period && ( props.canViewAnalytics || props.canViewInsights ) && (
 						<nav className='Hero__links'>
 							{ props.canViewAnalytics && (
-								<a href="index.php?page=altis-analytics">{ __( 'Analytics', 'altis-analytics' ) }</a>
+								<Link
+									url="index.php?page=altis-analytics"
+									location="dashboard"
+									linkText="Analytics"
```
roborourke, line 59: This is where the actual text for link is, so this where the translation functions can be used:
```suggestion
									linkText={ __( 'Analytics', 'altis-analytics' ) }
```

robindevitt:
> Something seems a little strange here, there are unrelated changes from another PR. What was the reason for force pushing?

aaah! I had conflicts with master, ran a merge and had all the commits come through and a lot of unrelated changes. I managed to undo the merge and then got the branch back to the [Update for WIP.](https://github.com/humanmade/aws-analytics/pull/415/commits/1e6387b48c3bbdeb8a2ad6f8eb3c89070d2ddd94) commit. I was then trying to get the branch back to that commit before trying the merge again.

src/blocks/ui/insights/personalization.js
```diff
@@ -21,15 +22,27 @@ const Personalization = ( {
...
+			start,
+			end,
+		} );
+	}, [ clientId, days, start ] );
 	const lift = useSelect( select => {
-		const current = select( 'analytics/xbs' ).getViews( clientId, { days: 7 } );
+		const current = select( 'analytics/xbs' ).getViews( clientId, {
+			days: 'P7D',
+			start: 'P7D',
+			end: 'P7D',
```
roborourke, line 39: I'm a bit confused by this, the start and end values should be date strings rather than durations. What's your aim with adding these here in this way rather than deriving a start and end from the duration? We shouldn't have both days, and start/end values as they're representing the same thing. This is pretty confusing as-is.

inc/blocks/rest_api/namespace.php
```diff
@@ -46,13 +46,25 @@ function init() : void {
...
-					'default' => 7,
+					'type' => 'string',
+					'default' => 'P7D',
 				],
 				'offset' => [
 					'description' => __( 'Number of days to offset data by. For example 3 means get data from before 3 days ago.', 'altis-analytics' ),
-					'type' => 'number',
-					'default' => 0,
+					'type' => 'string',
+					'default' => 'P0D',
```
roborourke, line 55: A duration of 0 days doesn't seem valid here. Offset here was used as a number of days previous to get data for but if we're using `start` and `end` instead it becomes redundant. We want to remove the use of durations on the backend and just work with start and end dates instead.

robindevitt: Would we not be doing away with the `offset` and `days`?

roborourke:
We have the basics of the tracking here from the new updated dashboard so we're changing the purpose of this issue a bit to normalise the API endpoints - to that end I'll remove it from the beta release milestone as it's not a hard requirement for now.

inc/blocks/rest_api/namespace.php
roborourke: We would be yes - but specifically on the backend, in the JS we want to standardise on using the period / duration notation.

I need to more fully review where things are at in terms of the API endpoints and how they're used - it's more involved than I first thought when I asked you to work on it.

roborourke:
Lets put this on hold for now until I've written up a proper spec for how everything should work in terms of the API endpoints.

PR #418, opened 2022-06-21
Allow granular usage of plugin features

roborourke:
This makes it possible to run the analytics plugin with certain features missing by virtue of not being in the distributed package.

Not ideal but seemed like the way to go if we want to avoid shipping all the code and allowing users to set a constant and get access to everything.

Related to https://github.com/humanmade/altis-accelerate/pull/11


shadyvb (approved):
Quite a big one! Couldn't trace all the changes tbh from old to new files, let me know if you expect that. 

Otherwise seems fine!

roborourke:
Yeah sorry, it's working well so far. I'll do some final checks today before merging.

PR #425, opened 2022-06-29
List table UI changes

shadyvb:
Ref humanmade/product-dev#1091


shadyvb:
I missed adding the font 🤦 , converting to draft to take care of that.

shadyvb:
Ah, I remember now that I needed to find an alternative font as SF Pro is only licensed to work on Apple products and cannot be used on the web.

src/accelerate/components/Dashboard.scss
```diff
@@ -228,44 +230,179 @@ body.index-php {
 
 .List {
 	position: relative;
-	font-size: 18px;
+	font-size: .875rem;
 	padding: 0 $sidebar-padding 40px;
+	-webkit-font-smoothing: antialiased;
+	-moz-osx-font-smoothing: grayscale;
```
roborourke, line 236: Are these necessary these days? I thought they could be detrimental to readability, at least when not using a web font. At any rate feels like something that would be handled more globally by WP itself.

src/accelerate/components/Dashboard.scss
```diff
@@ -303,13 +436,15 @@ body.index-php {
...
 		font-weight: 400;
-		border: none;
-		border-collapse: collapse;
-    	border-spacing: 0;
-		border-bottom: 1px solid #eee;
 
-		tr:nth-child(even) {
-			background: #f9f9f9;
+		border-collapse: separate;
+	    border-spacing: 0 1rem;
```
roborourke, line 441:
```suggestion
		border-spacing: 0 1rem;
```

src/accelerate/components/Dashboard.scss
```diff
@@ -328,16 +463,52 @@ body.index-php {
...
+			width: 110px;
+			padding: 0;
+			padding-right: 40px;
 
-		.table-th-links {
-			min-width: 80px;
+			img {
+				width: 105px;
+				height: 47px;
+				object-fit: cover;
```
roborourke, line 474: Thinking `object-position: center top` as well? You should should remove the width & height from here as it'll screw up the aspect ratio - they belong on the container element this object needs to fit into.

src/accelerate/components/List.tsx
```diff
@@ -1,31 +1,36 @@
...
 };
 
-export default function List( props: Props ) {
+export default function List ( props: Props ) {
 	// Filters.
-	const [ search, setSearch ] = useState<string>( '' );
-	const [ page, setPage ] = useState<number>( 1 );
-	const [ type, setType ] = useState<string|null>( null );
-	const [ user, setUser ] = useState<number|null>( null );
+	const [customFilter, setCustomFilter] = useState<string>( 'all' );
```
roborourke, line 23: Not sure what happened with the spacing here, not urgent but maybe check auto formatting settings.

src/accelerate/components/List.tsx
```diff
@@ -97,16 +131,39 @@ export default function List( props: Props ) {
...
+									className='dashicons-before dashicons-plus'
+								>
+									{ __( 'Add New', 'altis' ) }
+								</Button>
+							) }
+							renderContent={ ( { onClose } ) => (
+								<MenuGroup>
+									{/* <MenuItem onClick={ onAddNew( 'block' ) }>
+										{ __( 'Block', 'altis' ) }
+									</MenuItem> */}
```
roborourke, line 153: Why is this commented out? Might be good to expose if it helps drive the kind of activity we want to see?

src/accelerate/components/List.tsx
```diff
@@ -97,16 +131,39 @@ export default function List( props: Props ) {
...
+									<MenuItem onClick={ onAddNew( 'post' ) }>
+										{ __( 'Post', 'altis' ) }
+									</MenuItem>
```
roborourke, line 156: Not sure I follow why these are using `onClick` and not just regular `<a>` tags

src/accelerate/components/List.tsx
```diff
@@ -129,27 +186,38 @@ export default function List( props: Props ) {
 							}
 
 							return (
-								<tr key={ post.id }>
-									<td className="record-traffic">
-										<div className="traffic-bar" style={ { right: `${ 100 - ( post.views / maxViews * 100 ) }%` } }></div>{ compactMetric( post.views ) }
+								<tr key={post.id}>
+									<td className='record-thumbnail'>
+										{ post.thumbnail && (
+											<img src={ post.thumbnail } alt={ post.title } width="100" height="50" />
```
roborourke, line 192: This width & height are different to what was in the CSS but should be removed as we need to support various aspect ratios. Use a wrapper div that has a specific width & height then `object-fit` will work properly.

src/accelerate/components/List.tsx
```diff
@@ -129,27 +186,38 @@ export default function List( props: Props ) {
...
 									<td className={ `record-lift score-${ lift && lift >= 0 ? 'pos' : 'neg' }` }>
 										{ !! lift && ! isNaN( lift ) && ( lift >= 0 ? '↑' : '↓' ) }
 										{ !! lift && ! isNaN( lift ) && compactMetric( parseFloat( lift.toFixed( 1 ) ) ) }
 									</td>
-									<td className="record-author">
-										<img alt="" className="record-avatar" src={ post.author.avatar } />
-										{ post.author.name }
-									</td>
-									<td className="record-links">
-										{ post.url && ( <a href={ post.url }>{ __( 'View', 'altis' ) }</a> ) }
```
roborourke, line 151: Why remove the view link? I thought those were on the design still. If not it's ok to question what's on the designs if your instincts are telling you it makes sense to keep them.

src/accelerate/components/List.tsx
```diff
@@ -97,16 +131,39 @@ export default function List( props: Props ) {
...
+							) }
+							renderContent={ ( { onClose } ) => (
+								<MenuGroup>
+									{/* <MenuItem onClick={ onAddNew( 'block' ) }>
+										{ __( 'Block', 'altis' ) }
+									</MenuItem> */}
+									<MenuItem onClick={ onAddNew( 'post' ) }>
+										{ __( 'Post', 'altis' ) }
+									</MenuItem>
+									<MenuItem onClick={ onAddNew( 'page' ) }>
```
roborourke, line 157: Can we list all post types rather than hard coding? Not sure if there was any discussion / annotation to that effect. cc @noeltock

roborourke (changes requested):

inc/api/namespace.php
```diff
@@ -626,6 +626,14 @@ function get_top_data( $start, $end, ?Filter $filter = null ) {
 	$query = new WP_Query( $query_args );
 
 	foreach ( $query->posts as $i => $post ) {
+		$thumbnail_id = 0;
+		if ( $post->post_type === 'attachment' ) {
+			$thumbnail_id = $post->ID;
+		} else {
+			$thumbnail_id = get_post_thumbnail_id( $post ) ?: 0;
+		}
+		$thumbnail = $thumbnail_id ? wp_get_attachment_image_url( $thumbnail_id, '100x50' ) : '';
```
roborourke, line 635: Where is this image size coming from? I think we may be best off using the default `thumbnail` or `post-thumbnail` image size if it exists.

src/accelerate/components/Dashboard.scss
shadyvb: Dropping it to keep consistency with WP styles.

```suggestion
```

src/accelerate/components/Dashboard.scss
shadyvb: `object-position` is a nice addition, the width/height does not screw up the aspect ratio as the `object-fit` property handles positioning the actual image properly.
```suggestion
				object-fit: cover;
				object-position: center top;
```

src/accelerate/components/List.tsx
shadyvb: I don't think you can actually create a _reusable block_ right now ?

src/accelerate/components/List.tsx
shadyvb: Yeah, sorry about that. Will adjust my editor settings.

src/accelerate/components/List.tsx
shadyvb: Just using core components and that's how it works. Because it's not always navigating away. Happy to drop and check alternatives, it just seemed like a time saver.

src/accelerate/components/List.tsx
shadyvb: We didn't have any conversation about which types to allow, but I guess we can detect types and populate the list accordingly 👍

roborourke: Already being passed to the JS via localise script

src/accelerate/components/List.tsx
roborourke: Hmm, if it's a required prop then I guess it makes sense but it doesn't seem particularly accessible to me

src/accelerate/components/List.tsx
shadyvb: Yeah, was just using it, and it does seem you can create a reusable block via the URL.

roborourke: You can in Altis so might be because of the feature to show blocks in the admin. The arguments used when the post type is registered might need to be tweaked / filtered to allow it but `?post-new / edit.php?post_type=wp_block` should work.

src/accelerate/components/List.tsx
roborourke: The post type is `wp_block`, maybe that's why

src/accelerate/components/List.tsx
shadyvb: `object-fit` only applies to replaced elements like `` rather than their containers, and it assumes the element has an explicit width/height so the fitting would work. Tried a couple of scenarios and this was the best fit that worked as expected. Reference: https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit

src/accelerate/components/List.tsx
shadyvb: They were not on the designs, and the view action can be triggered by clicking the post link, which I assume the design included, it's also what I'd expect as a user.

image

shadyvb:
Addressed feedback, merging.

PR #435, opened 2022-07-01
Add inline spark charts to content explorer

roborourke:
Starts off https://github.com/humanmade/product-dev/issues/1093

Screenshot 2022-07-01 at 16 02 26

### DIFF
--- a/inc/api/namespace.php
+++ b/inc/api/namespace.php
@@ -298,8 +298,8 @@ function get_graph_data( $start, $end, $resolution = 'day', ?Filter $filter = nu
 				'field' => 'event_timestamp',
 				'interval' => $resolution,
 				'extended_bounds' => [
-					'min' => sprintf( '%d000', $start ),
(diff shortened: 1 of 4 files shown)


joehoyle:
This is what I see, but I think it's because I have very little data-- it does raise the question though: when someone first activates / the first day, the chart looks a bit wonky:

roborourke:
Not sure why but the histogram buckets are getting an extra day - I need to re-read the docs to understand, might need to remove a half day either end of the date range or something.

roborourke:
Nah, not sure, something wrong with the scale / domain calculations

shadyvb:
@joehoyle fixed the chart issue, not entirely sure if i did it correctly, the screenshot in the original post shows bars flowing bottom to top, while the code seemed to do it the reverse way, bit confused. But it seems to work now, perhaps Rob missed pushing a commit or something, or i got something wrong.

Also added titles to bars, to show the view count and day in text when hovering, and updated the bar color to match designs.

PR #440, opened 2022-07-06
Block preview API implementation

shadyvb:
ref https://github.com/humanmade/product-dev/issues/1094


inc/api/namespace.php
```diff
@@ -671,6 +671,26 @@ function get_top_data( $start, $end, ?Filter $filter = null ) {
 
 		$thumbnail = $thumbnail_id ? wp_get_attachment_image_url( $thumbnail_id, get_available_thumbnail_size() ) : '';
 
+		if ( $post->post_type === 'wp_block' ) {
+			$preview_url = sprintf(
+				get_home_url() . '?preview-block-id=%d&nonce=%s',
+				$post->ID,
+				wp_create_nonce( 'preview-block' )
```
joehoyle, line 678: Could we make the nonce specific to the post ID, so it's not a shared nonce for everything?

Also we need to make sure the current user has access to view the block, do we need a cap check for that?

joehoyle: Actually, I just thought... this needs a bit more consideration on how to handle private content. The issue with the nonce is it will change every 12 hours, when we don't probably want to change the URL to the `block-image` endpoint (to keep cache working.) So, I think we might need to us a hmac or something.

If we do that, we need to consider what we should do for users who don't have access to read the wp_block. I'm out for the day now but I think this logic might need to change a bit to accomadate.

shadyvb: I'm thinking the trade off between caching for 12 hours only and managing access to content would probably settle for the earlier, users can be demoted/removed, and this is a way to make sure they still have access to view content.

I'm assuming whomever have access to the dashboard, currently those with `edit_posts`, will have access to all content by default.

shadyvb: What I did forget to implement is adding a filter to manage disabling the feature, and maybe an additional cap check to allow for filtering private content manually.

shadyvb: Added in https://github.com/humanmade/aws-analytics/pull/440/commits/bdd5d712b870d41730bf45c042f557a0171b7c64

joehoyle: @shadyvb I think we need to cache for longer than 12 hours, as generating the previews is not going to be that fast.

I think fundamentally though, we need to make it so users can only view the previews if they have access to `read_post` for each post, so we should be wrapping it in checks for that.

So, I think we should:

- use a hmac for the key, instead of a nonce. A nonce is time based which I think doesn't make sense in this context. The hmac should be checked on the request to render the preview URL.
- Wrap the URL generation / image in a `current_user_can( 'read_post', $id )` to only show previews if the user has access to read that post.

Does that make sense?

shadyvb: Does that solve the issue where links are saved/cached/leaked ? If I understood correctly, the checks we need will be done while generating the links not while hitting them.

shadyvb: How about caching the previews in WP ? This way the preview is cached as needed, and we use WordPress safeguards to protect content. Thoughts ?

To clarify, WordPress will then request the preview URL and deliver the result to the authenticated user.

inc/api/namespace.php
```diff
@@ -671,6 +671,26 @@ function get_top_data( $start, $end, ?Filter $filter = null ) {
 
 		$thumbnail = $thumbnail_id ? wp_get_attachment_image_url( $thumbnail_id, get_available_thumbnail_size() ) : '';
 
+		if ( $post->post_type === 'wp_block' && apply_filters( 'altis.accelerate.allow_block_thumbnails', true, $post->ID ) ) {
+			$preview_url = sprintf(
+				get_home_url() . '?preview-block-id=%d&nonce=%s',
```
roborourke, line 676: Can probably put `get_home_url()` in the list of replacements rather than concatenate this bit?

inc/dashboard/namespace.php
```diff
@@ -144,3 +146,82 @@ function get_plugin_version() : string {
...
+}
+
+/**
+ * Return whether block thumbnail functionality is allowed.
+ *
+ * @param int $block_id Block ID.
+ *
+ * @return boolean
+ */
+function is_block_thumbnails_allowed( int $block_id = null ) : bool {
```
roborourke, line 190: Grammatically the non-plural version reads better as we're just checking a single block.

roborourke:
```suggestion
function is_block_thumbnail_allowed( int $block_id = null ) : bool {
```

inc/api/namespace.php
```diff
@@ -713,6 +714,27 @@ function get_top_data( $start, $end, ?Filter $filter = null ) {
 
 		$thumbnail = $thumbnail_id ? wp_get_attachment_image_url( $thumbnail_id, get_available_thumbnail_size() ) : '';
 
+		if ( $post->post_type === 'wp_block' && Dashboard\is_block_thumbnails_allowed( $post->ID ) ) {
```
roborourke, line 717:
```suggestion
		// Get block thumbnail from screen grab API.
		if ( $post->post_type === 'wp_block' && Dashboard\is_block_thumbnail_allowed( $post->ID ) ) {
```

roborourke: Should the block thumbnail allowed function check the post type is valid too? I don't think it's vital but would avoid the need for checking it here and make the function more specific. Then again, maybe we will use this API even for regular posts that don't support featured images or don't have them in future.

inc/dashboard/namespace.php
```diff
@@ -144,3 +146,82 @@ function get_plugin_version() : string {
...
+		|| $hmac !== get_block_thumbnail_request_hmac( $block_id )
+	) {
+		return;
+	}
+
+	$allow_block_thumbnails = is_block_thumbnails_allowed( $block_id );
+
+	if ( ! $allow_block_thumbnails ) {
+		return;
+	}
```
roborourke, line 174:
```suggestion
	if ( ! is_block_thumbnail_allowed( $block_id ) ) {
		return;
	}
```

roborourke (approved):
Couple of suggestions that are optional around naming etc..

Lets see what the performance is like - if the request is blocking I'm thinking we may want to generate the thumbnails by triggering the API requests on a background task whenever a block post is saved. Shall we do that in a follow up? It doesn't change this code at any rate

roborourke:
Derp, request will happen client side when the image is requested. Ignore me.

inc/api/namespace.php
```diff
@@ -713,6 +714,28 @@ function get_top_data( $start, $end, ?Filter $filter = null ) {
...
+				Dashboard\get_block_thumbnail_request_hmac( $post->ID )
+			);
+
+			$version = strtotime( $post->post_modified_gmt );
+
+			$thumbnail = add_query_arg(
+				[
+					'url' => urlencode( $preview_url ),
+					'width' => 100,
+					'selector' => 'altis-block-preview',
```
joehoyle, line 732: Ahh this needs to be `.altis-block-preview` (a css selector)

inc/dashboard/namespace.php
```diff
@@ -144,3 +146,80 @@ function get_plugin_version() : string {
...
+/**
+ * Override template to output a block preview markup.
+ *
+ * @return void
+ */
+function block_thumbnail_template_override() : void {
+	global $content_width;
+
+	get_header();
+	printf( '<style>.altis-block-preview { width: %spx; margin: 0 auto; }</style>', (int) $content_width );
```
joehoyle, line 217: When the theme doesn't set `content_width` this will show `0`, which then means the block doesn't show _anything_

PR #443, opened 2022-07-14
Remove dependency on wp-component to fix webpack prod build

shadyvb:


roborourke:
Superceded by #446 - adding wp-components css to dev mode instead

PR #452, opened 2022-07-19
Multisite compat & bug fixes

roborourke:
- Ensure AUTH_KEY is defined or read from site options / options
- Ensure correct namespace for accelerate version
- Allow preview thumbs for xb post type as well, subset of reusable blocks

### DIFF
--- a/inc/dashboard/namespace.php
+++ b/inc/dashboard/namespace.php
@@ -77,16 +77,15 @@ function load_dashboard() {
 	$title = __( 'Accelerate Dashboard', 'altis' );
 	$user = wp_get_current_user();
 
-	$post_types = array_merge(
-		// Trackable post types that do not have their own front end URL.
-		[
-			get_post_type_object( 'wp_block' ),
-		],
-		get_post_types( [
-			'show_in_menu' => true,
-			'public' => true,
-		], 'objects' )
-	);
+	$post_types = get_post_types( [
+		'show_in_menu' => true,
+		'public' => true,
+	], 'objects' );
+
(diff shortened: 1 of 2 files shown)


inc/dashboard/namespace.php
```diff
@@ -146,7 +145,7 @@ function render_page() {
 function get_plugin_version() : string {
 	// Only show version if this is embedded in the accelerate plugin.
 	if ( defined( 'Altis\\Accelerate\\VERSION' ) ) {
-		return Altis\Accelerate\VERSION;
+		return \Altis\Accelerate\VERSION;
 	}
```
robindevitt, line 149: Why would the `return` be "`\Altis\Accelerate\VERSION;`" but the `if` statement is "`Altis\\Accelerate\\VERSION`"

shadyvb (approved):

robindevitt (commented):
Just a single question from my based on the one change

inc/dashboard/namespace.php
shadyvb: Should be `Altis\Accelerate\VERSION` i think, I missed that, good shout!

roborourke: In the if statement it's escaped using double backslashes as we're checking for the existence of a constant by name as a string. In that instance the `defined()` function isn't scoped to the current namespace. Adding the preceding backslash before the `return` statement is just to make absolutely certain we're reading from the root of the namespaces.

robindevitt: Thanks @roborourke

PR #458, opened 2022-07-21
Admin notices wrapper and style.

robindevitt:
Related to https://github.com/humanmade/product-dev/issues/1118

Added a general class for the wrapper and the CSS targets any child elements to the parent.

#### Before Screenshot
first

#### After Screenshot
second

### DIFF
--- a/inc/dashboard/namespace.php
+++ b/inc/dashboard/namespace.php
@@ -26,6 +26,8 @@ function setup() {
 
 	add_action( 'load-index.php', __NAMESPACE__ . '\\load_dashboard' );
 	add_action( 'admin_menu', __NAMESPACE__ . '\\add_widgets_submenu' );
+	add_action( 'admin_notices', __NAMESPACE__ . '\\add_notices_wrapper_open', 0 );
+	add_action( 'admin_notices', __NAMESPACE__ . '\\add_notices_wrapper_close', 999 );
 
 	add_action( 'pre_get_posts', __NAMESPACE__ . '\\block_preview_check' );
 }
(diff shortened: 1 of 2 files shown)


robindevitt:
Potentially we need include a way to make sure the div is only added on the Accelerate Dashboard and not on other pages?

src/accelerate/components/App.css
```diff
@@ -20,3 +20,12 @@
...
 	grid-template-columns: repeat( 3, 1fr );
 }
+
+.notices__wrap {
+	background-color: #f5f6f8;
+	overflow: hidden;
+}
+
+.notices__wrap > * {
+	margin: 15px 5% 0 5% !important;
```
roborourke, line 30: Do we need `!important` here? It's best to avoid whenever possible. Can we rely on CSS specificity instead e.g. by adding another class name or ID to the rule like `#wpbody .notices__wrap > *`?

Also default notice styling appears to have a 5px bottom margin too:

```suggestion
	margin: 15px 5% 5px 5%;
```

inc/dashboard/namespace.php
```diff
@@ -26,6 +26,8 @@ function setup() {
 
 	add_action( 'load-index.php', __NAMESPACE__ . '\\load_dashboard' );
 	add_action( 'admin_menu', __NAMESPACE__ . '\\add_widgets_submenu' );
+	add_action( 'admin_notices', __NAMESPACE__ . '\\add_notices_wrapper_open', 0 );
+	add_action( 'admin_notices', __NAMESPACE__ . '\\add_notices_wrapper_close', 999 );
```
roborourke, line 30: We could probably go higher on the closing div priority, but it's kind of hit and miss whether someone will use a reasonable value though. Lets see how we go.

In answer to your question about only showing the wrapper on the Altis dashboard page we do need to do that. You can place the `add_action()` calls inside the `load_dashboard()` function after the early returns.

roborourke (dismissed):

shadyvb (approved):

PR #466, opened 2022-07-28
Cache analytics requests and use local shards

roborourke:
Adds performance boosting query parameters to searches and reduces the largest terms aggregations to more manageable sizes for high traffic sites.


joehoyle (approved):
Looks sane -- were you able to test the query in production to see if it is indeed faster

roborourke:
@joehoyle yeah, just with Kibana directly but it went from a first run of 10s or so down to ~300ms on subsequent runs, even lower for some of the other audience field related queries down to ~20ms on subsequent runs.

PR #470, opened 2022-08-02
Add image loading component for block previews

roborourke:
Displays the content loader svg until the image is loaded.

Fixes https://github.com/humanmade/product-dev/issues/1116

### DIFF
--- a/src/accelerate/components/Image.tsx
+++ b/src/accelerate/components/Image.tsx
@@ -1,0 +1,43 @@
+import React, { useState } from 'react';
+import ContentLoader from 'react-content-loader';
+
+type Props = {
+	alt?: string,
+	className?: string,
+	src: string,
+	width?: number,
+	height?: number,
+};
+
+const loaderProps = {
+	speed: 2,
+	foregroundColor: "#f5f6f8",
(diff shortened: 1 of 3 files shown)


src/accelerate/components/Image.tsx
```diff
@@ -0,0 +1,43 @@
...
+	speed: 2,
+	foregroundColor: "#f5f6f8",
+	backgroundColor: "#fff",
+};
+
+export default function Image( props: Props ) {
+	const [ isLoaded, setIsLoaded ] = useState( false );
+	return (
+		<>
+			<ContentLoader
```
joehoyle, line 22: Is there a reason we don't wrap this in a conditional rather than `display: none` when it's not needed?

roborourke: 🤔  not necessarily, we could do. I was thinking of having some fade in transition by layering the image over the loader but it was getting a bit fiddly

joehoyle: Ahh ok, that's fine then yeah if we want a transition now / in the future.

joehoyle (approved):

PR #478, opened 2022-08-03
Improve audience UI performance

roborourke:
- Improve cache hit rate by clamping to nearest day
- Always return field data even if ES fails so UI is not blocked
- Reduce terms query for possible values to top 20
- Filter fields query to pageViews rather than all events
- Fix top hits percentage calculation


roborourke:
If this is approved can you merge and then merge the backports and create new releases from the release branches backported to?

shadyvb (approved):

PR #483, opened 2022-08-10
CE - Various design changes

noeltock:
Work in progress

### DIFF
--- a/inc/api/namespace.php
+++ b/inc/api/namespace.php
@@ -49,7 +49,7 @@ function register_endpoints() {
 		],
 		'interval' => [
 			'type' => 'string',
-			'default' => 'day',
+			'default' => '1d',
 		],
 	];
 
@@ -81,6 +81,20 @@ function register_endpoints() {
 			'page' => [
 				'type' => 'number',
 				'default' => 1,
+			],
+		], $date_args ),
+	] );
+
+	register_rest_route( API_NAMESPACE, 'diff', [
+		'method' => WP_REST_Server::READABLE,
+		'callback' => __NAMESPACE__ . '\\get_diff',
+		'permission_callback' => __NAMESPACE__ . '\\check_analytics_permission',
+		'args' => array_merge( [
+			'ids' => [
+				'type' => 'array',
+				'items' => [
+					'type' => 'number',
+				],
 			],
 		], $date_args ),
 	] );
@@ -153,6 +167,20 @@ function get_top( WP_REST_Request $request ) {
 	] );
 
 	return $response;
+}
+
+/**
+ * Handle diff endpoint.
+ *
+ * @param WP_REST_Request $request The request object.
+ * @return WP_Error|array
+ */
+function get_diff( WP_REST_Request $request ) {
+	$start = strtotime( $request['start'] );
+	$end = strtotime( $request['end'] );
+	$post_ids = $request['ids'] ?? [];
+
+	return get_post_diff_data( $post_ids, $start, $end, $request['interval'] );
 }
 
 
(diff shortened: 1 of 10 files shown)


noeltock:
Hey @roborourke , thanks for offering to support on these pieces, from the original design, open to ideas:

- Revert colours to Altis branded ones per design
- Narrower bars for spark barchart. Line 27 `width = 160`, I propose variable widths on date range selected:
  - 7D `100` 
  - 30D `120` 
  - 90D `150` 
- Not sure what happened to the original comparison % for 7D and 30Ds, along the lines of `↑ 2%` or `↓ 2%`, probably makes sense to stick with rounded integers. Is that too far out of scope, or hard to do, maybe that's why it was left out? 
- On the meta for posts (above the name), we still have the ability to show:
  - Posts: `17D OLD`, if under 1 day `1H OLD`, etc.
  - Pages: `PARENT NAME` if one exists
- Dropdown alignment.  

Screenshot 2022-08-17 at 11 34 45

Screenshot 2022-08-17 at 11 34 52

Let me know if there are any questions, don't have anything to push up from my branch.. cheers!

roborourke:
Screenshot 2022-08-17 at 15 54 52

The relative date display in the format you have on the mockup is going to be a pain to implement, this is using moment.js's built-in `fromNow()` method which is internationalised through WP. Happy with that?

roborourke:
> Not sure what happened to the original comparison % for 7D and 30Ds, along the lines of ↑ 2% or ↓ 2%, probably makes sense to stick with rounded integers. Is that too far out of scope, or hard to do, maybe that's why it was left out?

I think it was descoped, but to be honest to fetch this data is going to be quite an expensive query. I think I might have to split it out into per post queries. Individually they should be fast.

roborourke:
Progress so far:

Screenshot 2022-08-17 at 16 33 31

There's not much to indicate what the parent page link is besides colour, but especially for XBs it's useful to have it there. Let me know if you have any further thoughts.

roborourke:
This is the contrast check output for the small grey text on the gray background https://webaim.org/resources/contrastchecker/?fcolor=7B7B7B&bcolor=F5F6F8

It will pass if the grey text is `#707070`

noeltock:
> The relative date display in the format you have on the mockup is going to be a pain to implement, this is using moment.js's built-in fromNow() method which is internationalised through WP. Happy with that?

Yep! I'm assuming you saw https://github.com/moment/moment/issues/2781#issuecomment-160739129 ?

> There's not much to indicate what the parent page link is besides colour, but especially for XBs it's useful to have it there. Let me know if you have any further thoughts.

With the shift/movement around global blocks as the common denominator, does it still make sense to show an arbitrary page in which an XB is located? Or does it show only show if the XB has a single child page (in the global block/XB world, I guess blocks don't have pages as parents, but as children, at least conceptually)? Thoughts?

roborourke:
> With the shift/movement around global blocks as the common denominator, does it still make sense to show an arbitrary page in which an XB is located? Or does it show only show if the XB has a single child page (in the global block/XB world, I guess blocks don't have pages as parents, but as children, at least conceptually)? Thoughts?

From a technical standpoint an XB can't exist in isolation, it has to be created somewhere whether that's a global block, post, page etc... so it's not arbitrary in that sense. I didn't think it made much sense initially, especially with things like A/B tests that you could create the block but not be gathering any data for it. We can maybe revisit that a bit with some work, e.g. a "waiting for data" status to know if it's used somewhere.

If an XB is on a post or page it'll show that as the parent, for an XB on a global block the global block will show as the parent right now. It was too complicated to build or modify the reusable blocks interface at the time I originally coded them.

Regarding XBs/Global blocks as having child pages - that makes sense in terms of there being a one-to-many relationship but conceptually it's a bit different. They might be on template parts like header / footer, or broadcasts, or whatever. We'd need to think of all of those elements as content pieces. I'm not really sure if people will really get into that mode of thinking or look at it in that way with WP, even in light of FSE. Some stuff will be in the db, some will be in code templates.

There was a question early on about whether we should require XBs to be only allowed in reusable blocks but we decided they should be immediately usable. If we can figure out a way to make it as simple to start an XB as it is now without explicitly creating a global block first we could try and see if it's any easier now.

PR #492, opened 2022-08-25
Improve scheduled task performance

### LINKED ISSUE #490: Avoid scheduling events on every request
Moved from humanmade/Cavalcade#116 per @kovshenin's [comment](https://github.com/humanmade/Cavalcade/issues/116#issuecomment-1224886054) on that ticket.

When viewing any page on a site where AWS Analytics is enabled, a number of Cavalcade queries are made and a variety of in-memory cache values are set. Per the trace in humanmade/Cavalcade#116, the origin of these queries is this function in AWS Analytics:

https://github.com/humanmade/aws-analytics/blob/aa5763c1df8e13ecb457b9db9a8deddded8ccd63/inc/namespace.php#L81-L89

Is there a reason these requests need to be made on every request, hooked to `init`, rather than in their own cron entry? While each individual request is not significant, making several needless queries and taking time to stash the data in-memory, if not necessary to render the page, makes every pageview in Altis sites slightly slower.

roborourke:
Avoids unnecessary db lookups by using persistent cache when available, also defers scheduling to `admin_footer` action where possible. For AB tests we check if its in the admin or not before scheduling the task.

Fixes #490

### DIFF
--- a/inc/experiments/namespace.php
+++ b/inc/experiments/namespace.php
@@ -402,11 +402,9 @@ function register_post_ab_test( string $test_id, array $options ) {
 	if (
 		( ! defined( 'WP_INSTALLING' ) || ! WP_INSTALLING ) &&
 		is_admin() &&
-		! wp_cache_get( "{$test_id}_cron", 'altis.analytics' ) &&
 		! wp_next_scheduled( 'altis_post_ab_test_cron', [ $test_id ] )
 	) {
 		wp_schedule_event( time(), 'hourly', 'altis_post_ab_test_cron', [ $test_id ] );
-		wp_cache_set( "{$test_id}_cron", 1, 'altis.analytics', HOUR_IN_SECONDS );
 	}
 
 	/**
--- a/inc/experiments/namespace.php
+++ b/inc/experiments/namespace.php
@@ -399,8 +399,14 @@ function register_post_ab_test( string $test_id, array $options ) {
 	add_action( "altis.experiments.test.winner_found.{$test_id}", $options['winner_callback'], 10, 2 );
 
 	// Set up background task.
-	if ( ( ! defined( 'WP_INSTALLING' ) || ! WP_INSTALLING ) && ! wp_next_scheduled( 'altis_post_ab_test_cron', [ $test_id ] ) ) {
+	if (
+		( ! defined( 'WP_INSTALLING' ) || ! WP_INSTALLING ) &&
+		is_admin() &&
+		! wp_cache_get( "{$test_id}_cron", 'altis.analytics' ) &&
+		! wp_next_scheduled( 'altis_post_ab_test_cron', [ $test_id ] )
+	) {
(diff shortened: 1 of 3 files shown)


kovshenin:
Not sure I like the use of persistent object cache here. I think moving the Cavalcade schedule reads to the admin is more than enough to fix the outlined problem. The additonal caching layer feels unnecessary and also slightly innefficient -- these low-frequency keys will be the first on their way out should Redis need to evict items.

If we wanted to iterate on the approach in general I think we should consider scheduling these events as part of an upgrade routine, storing a plugin version in an auto-loaded entry in `wp_options` and performing an upgrade only if the stored version is lower than the current version. Fairly similar to how Core does it. That way we'll have no extra reads or writes to the DB or Redis, nor would we need to check for these schedules "every once in a while", even if limited to wp-admin. The only downside is a few bytes added to `alloptions`.

```php
add_action( 'admin_footer', function() {
    $installed_version = get_option( 'aws_analytics.version' );
    if ( version_compare( $installed_version, CURRENT_VERSION, '<' ) ) {
        update_option( 'aws_analytics.version', CURRENT_VERSION );
        check_and_schedule_events();
    }
} );
```

What do you think?

roborourke:
Yeah I'd thought about the option version but the versions aren't always updated properly, removed the persistent stuff for now.

kadamwhite (approved):
Thanks for this, @roborourke

kovshenin (approved):

inc/experiments/namespace.php
```diff
@@ -399,7 +399,11 @@ function register_post_ab_test( string $test_id, array $options ) {
 	add_action( "altis.experiments.test.winner_found.{$test_id}", $options['winner_callback'], 10, 2 );
 
 	// Set up background task.
-	if ( ( ! defined( 'WP_INSTALLING' ) || ! WP_INSTALLING ) && ! wp_next_scheduled( 'altis_post_ab_test_cron', [ $test_id ] ) ) {
+	if (
+		( ! defined( 'WP_INSTALLING' ) || ! WP_INSTALLING ) &&
+		is_admin() &&
```
kovshenin, line 404: @roborourke I think the `is_admin()` can potentially make this portion of the code never run. Do you know when and how register_post_ab_test() is invoked? Does it happen based on a user action? If it does then I think that's already low-frequency enough and doesn't need any further optimization.

kovshenin: Might have jumped the gun with my approval, sorry.

roborourke: It has to be called like `register_post_type()` on the `init` hook

kovshenin: Ah ok, seems like it runs both in and outside of wp-admin, so should be ok I think:

https://github.com/humanmade/aws-analytics/blob/66bafa0e93dbd015ddf383c40ac46c2e0f8b303f/inc/experiments/titles/namespace.php#L18-L30

PR #499, opened 2022-08-26
Turn off linting in third-party polyfill

kadamwhite:
This appears to be causing a consistent HM Linter failure.


kadamwhite:
Well, guess that wasn't the only issue 😆 but now it'll be easier to see PHP style issues.

PR #508, opened 2022-09-19
Add a `display` element for the cron schedule

johnbillion:
Custom cron schedules should have a `display` element in the array which is used by cron schedule management plugins. This adds it using [the same format as that found in altis-cloud](https://github.com/humanmade/altis-cloud/blob/1c451030c8d01cdd4c5f72914203e5e6393a2d17/inc/healthcheck/cavalcade/namespace.php#L54).

### DIFF
--- a/inc/export/cron/namespace.php
+++ b/inc/export/cron/namespace.php
@@ -90,6 +90,7 @@ function setup_cron_interval( $schedules ) : array {
 	return array_merge( $schedules, [
 		ALTIS_ANALYTICS_EXPORT_CRON_FREQUENCY => [
 			'interval' => ALTIS_ANALYTICS_EXPORT_CRON_FREQUENCY_INTERVAL,
+			'display' => __( 'Altis Analytics Schedule (10 mins)', 'altis-analytics' ),
 		],
 	] );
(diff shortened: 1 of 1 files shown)


inc/export/cron/namespace.php
```diff
@@ -90,6 +90,7 @@ function setup_cron_interval( $schedules ) : array {
 	return array_merge( $schedules, [
 		ALTIS_ANALYTICS_EXPORT_CRON_FREQUENCY => [
 			'interval' => ALTIS_ANALYTICS_EXPORT_CRON_FREQUENCY_INTERVAL,
+			'display' => __( 'Altis Analytics Schedule (10 mins)', 'altis-analytics' ),
```
roborourke, line 93:
```suggestion
			'display' => __( 'Altis Analytics Export Schedule (10 mins)', 'altis-analytics' ),
```

roborourke (approved):

inc/export/cron/namespace.php
roborourke: We should probably make the schedule key more specific to this plugin too in this case.

PR #509, opened 2022-09-19
Convert experience block queries to use ClickHouse

roborourke:
Fixes https://github.com/humanmade/altis-accelerate/issues/71


inc/blocks/rest_api/namespace.php
```diff
@@ -156,7 +152,7 @@ function sanitize_id( $param ) : string {
  */
 function check_views_permission() : bool {
 	$type = get_post_type_object( Audiences\POST_TYPE );
-	return current_user_can( $type->cap->read );
+	return true; // current_user_can( $type->cap->read );
```
shadyvb, line 155: Is this a temporary hack?

shadyvb (approved):
LGTM

inc/blocks/rest_api/namespace.php
roborourke: 🤦‍♂️  yep, will sort it before merging

PR #510, opened 2022-09-21
Convert Dashboard API endpoints to use ClickHouse

roborourke:
Fixes https://github.com/humanmade/altis-accelerate/issues/70

### DIFF
--- a/inc/api/namespace.php
+++ b/inc/api/namespace.php
@@ -360,6 +360,10 @@ function get_graph_data( $start, $end, $resolution = 'day', ?Filter $filter = nu
 		$query['aggs'][ $name ] = $agg_options['aggregation'];
 	}
 
+	// Build SQL query.
+	$query = $wpdb->prepare();
+
+
 	$key = sprintf( 'analytics:stats:%s', sha1( serialize( $query ) ) );
 	$cache = wp_cache_get( $key, 'altis' );
 	if ( $cache ) {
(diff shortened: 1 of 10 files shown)


inc/api/namespace.php
```diff
@@ -1024,81 +809,84 @@ function get_post_diff_data( array $post_ids, $start, $end, $resolution = '1d' )
...
-			$res = Utils\query( $query, [
-				'preference' => __FUNCTION__,
-			] );
+	// Generate fields for histogram of current date range.
+	$buckets = [];
+	$bucket_start = $start;
+	while ( $bucket_start < $end ) {
+		$bucket_end = strtotime( "+{$resolution}", $bucket_start );
+		$buckets[ $bucket_start ] = sprintf(
+			'uniqCombined64If(endpoint_id, event_timestamp >= toDateTime64(%1$d,3) AND event_timestamp < toDateTime64(%2$d,3)) as `%1$d`',
```
shadyvb, line 838: This looks a bit weird, particularly the last `as` bit:
```
as `%1$d`
```

shadyvb (approved):
LGTM, with one little observation which might be okay.

inc/api/namespace.php
```diff
@@ -223,285 +205,146 @@ function get_query( array $events, int $start, int $end, array $aggs, ?Filter $f
...
+			count() as views,
+			-- uniques
+			uniqCombined64(endpoint_id) as uniques,
+			-- returning
+			uniqCombined64If(endpoint_id, endpoint_metrics['sessions'] > 1) as returning,
+			-- bounce
+			uniqCombined64If(endpoint_id, endpoint_metrics['pageViews'] < 2) as bounce
+		FROM analytics
+		WHERE
+			attributes['blogId'] = %s
```
rmccue, line 232: Does Clickhouse support parameterised queries? That'd be better than building a raw query

roborourke: Appears so, wasn't aware of that feature https://clickhouse.com/docs/en/interfaces/http/#cli-queries-with-parameters

inc/api/namespace.php
roborourke: I'm naming the field selected as the timestamp for the start of the histogram bucket, makes the post-processing very minimal.

inc/api/namespace.php
roborourke: @rmccue I'll do a follow up to parameterise the queries.

PR #512, opened 2022-09-22
Respect the cached value for audience config even if it's empty

johnbillion:
On a site with no audiences configured, Query Monitor reports a duplicate database query coming from the aws-analytics module. This is because the cached value for a site with no audience data is an empty array, therefore the cache check in `get_audience_config()` isn't satisfied because it evaluates to false.

```
> composer server cli -- cache get 'audiences' 'altis.analytics'
array (
)
```

By changing this to a check for `is_array()` the empty cached value is respected.

As a follow-up, the `Altis\Analytics\Preview\get_script_data()` function does not use any caching, but it probably should.


roborourke (approved):
Thanks!

PR #513, opened 2022-09-23
Use parameterised queries for ClickHouse

roborourke:
`$wpdb->prepare` is less appropriate and less powerful compared to ClickHouse built in parameterised value substitution.

This also normalises and corrects a few things such as date ranges, datetime types and the use of the `blog_id` column rather than the attribute as this is part of the sorting key for the table.

### DIFF
--- a/inc/api/namespace.php
+++ b/inc/api/namespace.php
@@ -195,8 +195,6 @@ function get_diff( WP_REST_Request $request ) {
  * @return array|WP_error
  */
 function get_graph_data( $start, $end, $resolution = '1 day', ?Filter $filter = null ) {
-	global $wpdb;
-
 	$query_where = [ '1=1' ];
 	$query_params = [
 		'blog_id' => get_current_blog_id(),
@@ -325,10 +323,14 @@ function get_graph_data( $start, $end, $resolution = '1 day', ?Filter $filter = 
 			GROUP BY {$field}
 			ORDER BY `value` DESC
 			LIMIT {limit:UInt8}",
-			array_merge( $query_params, [
-				'event_type' => $agg_options['event'],
-				'limit' => $agg_options['limit'],
-			], $agg_options['where_params'] ?? [] )
+			array_merge(
+				$query_params,
+				[
+					'event_type' => $agg_options['event'],
+					'limit' => $agg_options['limit'],
+				],
+				$agg_options['where_params'] ?? []
+			)
 		);
 
 		if ( is_wp_error( $res ) ) {
@@ -367,16 +369,17 @@ function get_top_data( $start, $end, ?Filter $filter = null ) {
 
 	if ( ! empty( $filter ) ) {
 		if ( $filter->type ) {
-			$types_where = array_map( function ( $type ) use ( $query_params ) {
+			$types = explode( ',', $filter->type );
+			$types_where = array_map( function ( $type, $idx ) use ( &$query_params ) {
 				if ( $type === 'xb' ) {
 					return "event_type IN ('experienceView', 'conversion')";
 				}
 				if ( $type === 'wp_block' ) {
 					return "event_type = 'blockView'";
 				}
-				$query_params[ 'type_' . $type ] = $type;
-				return sprintf( "event_type = 'pageView' AND attributes['postType'] = {type_%s:String}", $type );
-			}, explode( ',', $filter->type ) );
+				$query_params[ 'type_' . $idx ] = $type;
+				return sprintf( "event_type = 'pageView' AND attributes['postType'] = {type_%s:String}", $idx );
+			}, $types, array_keys( $types ) );
 			$query_where[] = sprintf( '(%s)', implode( ') OR (', $types_where ) );
 		}
 		if ( $filter->path ) {
--- a/inc/api/namespace.php
+++ b/inc/api/namespace.php
@@ -198,6 +198,11 @@ function get_graph_data( $start, $end, $resolution = '1 day', ?Filter $filter = 
 	global $wpdb;
 
 	$query_where = [ '1=1' ];
+	$query_params = [
+		'param_blog_id' => get_current_blog_id(),
+		'param_start' => (int) $start,
+		'param_end' => (int) $end,
+	];
 
 	if ( ! empty( $filter ) ) {
 		if ( $filter->path ) {
@@ -207,20 +212,18 @@ function get_graph_data( $start, $end, $resolution = '1 day', ?Filter $filter = 
 			} else {
 				$url_path_alt = 'http' . substr( $url_path, 5 );
 			}
-			$query_where[] = $wpdb->prepare(
-				"attributes['url'] = %s OR attributes['url'] = %s",
-				$url_path,
-				$url_path_alt
-			);
+			$query_where[] = "attributes['url'] = {path:String} OR attributes['url'] = {path_alt:String}";
+			$query_params['param_path'] = $url_path;
+			$query_params['param_path_alt'] = $url_path_alt;
 		}
 	}
 
 	$query_where = sprintf( 'AND (%s)', implode( ') AND (', $query_where ) );
 
 	// Build SQL query.
-	$query = $wpdb->prepare(
+	$query =
 		"SELECT
-			toUnixTimestamp(toStartOfInterval(event_timestamp, INTERVAL {$resolution})) as `date`,
+			toStartOfInterval(event_timestamp, INTERVAL {$resolution}) as `date`,
 			-- views
 			count() as views,
 			-- uniques
@@ -231,27 +234,24 @@ function get_graph_data( $start, $end, $resolution = '1 day', ?Filter $filter = 
 			uniqCombined64If(endpoint_id, endpoint_metrics['pageViews'] < 2) as bounce
 		FROM analytics
 		WHERE
-			attributes['blogId'] = %s
-			AND event_timestamp >= toDateTime64(intDiv(%d,1000),3)
-			AND event_timestamp <= toDateTime64(intDiv(%d,1000),3)
+			blog_id = {blog_id:String}
+			AND event_timestamp >= toDateTime64({start:UInt64},3)
+			AND event_timestamp <= toDateTime64({end:UInt64},3)
 			AND event_type = 'pageView'
 			{$query_where}
 		GROUP BY
 			`date`
 			WITH ROLLUP
-		ORDER BY `date` ASC",
-		get_current_blog_id(),
-		(int) sprintf( '%d000', $start ),
-		(int) sprintf( '%d999', $end )
-	);
+		ORDER BY `date` ASC
+			WITH FILL FROM toDateTime({start:UInt64}) STEP INTERVAL {$resolution}";
 
-	$key = sprintf( 'analytics:stats:%s', sha1( serialize( $query ) ) );
+	$key = sprintf( 'analytics:stats:%s', sha1( serialize( [ $query, $query_params ] ) ) );
 	$cache = wp_cache_get( $key, 'altis' );
 	if ( $cache ) {
 		return $cache;
 	}
 
-	$res = Utils\clickhouse_query( $query, '', 'array' );
+	$res = Utils\clickhouse_query( $query, $query_params );
 
 	if ( is_wp_error( $res ) ) {
 		return $res;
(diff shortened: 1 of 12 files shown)


inc/api/namespace.php
```diff
@@ -198,6 +198,11 @@ function get_graph_data( $start, $end, $resolution = '1 day', ?Filter $filter =
 	global $wpdb;
 
 	$query_where = [ '1=1' ];
+	$query_params = [
+		'param_blog_id' => get_current_blog_id(),
```
rmccue, line 202: This kind of leaks the REST API behaviour with the prefix included. I'd expect
```
Utils\clickhouse_query( 'SELECT * FROM analytics WHERE blog_id = {blog_id:string}', [
    'blog_id' => get_current_blog_id(),
] )
```

Suggest moving the prefix to the query function instead

inc/api/namespace.php
```diff
@@ -231,27 +234,24 @@ function get_graph_data( $start, $end, $resolution = '1 day', ?Filter $filter =
...
-		ORDER BY `date` ASC",
-		get_current_blog_id(),
-		(int) sprintf( '%d000', $start ),
-		(int) sprintf( '%d999', $end )
-	);
+		ORDER BY `date` ASC
+			WITH FILL FROM toDateTime({start:UInt64}) STEP INTERVAL {$resolution}";
 
-	$key = sprintf( 'analytics:stats:%s', sha1( serialize( $query ) ) );
+	$key = sprintf( 'analytics:stats:%s', sha1( serialize( [ $query, $query_params ] ) ) );
```
rmccue, line 248: This may be worth abstracting; `function key_for_query( string $query, ?array $params ) : string;`

inc/api/namespace.php
```diff
@@ -308,29 +307,30 @@ function get_graph_data( $start, $end, $resolution = '1 day', ?Filter $filter =
 			continue;
 		}
 
-		$query = $wpdb->prepare(
+		if ( ! empty(  $agg_options['where'] ) ) {
```
rmccue: Lint.

inc/experiments/namespace.php
```diff
@@ -974,16 +974,28 @@ function process_post_ab_test_result( string $test_id, int $post_id ) {
...
 	// Process event filter.
-	$query_filter = '1=1';
-	if ( is_callable( $test['query_filter'] ) ) {
-		$query_filter = call_user_func_array( $test['query_filter'], [ $post_id ] );
+	$query_filter = $test['query_filter'] ?: '1=1';
+	$query_params = [
+		'param_blog_id' => get_current_blog_id(),
+		'param_ts' => $data['timestamp'] ?? 0,
+		'param_view_event' => $test['view'] ?? 'testView',
+		'param_goal_event' => $goal[0] ?? '__none__', // Special fallback value - if no goal is set the test will never convert.
```
rmccue, line 982: Suggest making this magic value a constant

inc/utils/namespace.php
```diff
@@ -1094,14 +1094,13 @@ function get_s3_client( array $args = [] ) : ? S3Client {
...
  * @param string $query SQL statement, if $body is present it will be encoded into the request URL.
- * @param string $body Optional query body. For use with queries like INSERT with JsonEachRow format.
- * @param string $return Optional return type. Can be 'auto', 'array', 'object' or 'raw'.
+ * @param array $params Array of query parameters to add to the query string, use `param_` prefixed keys for interpolation of values.
+ * @param string $return Optional return type. Can be 'array', 'object' or 'raw', default 'array'.
+ * @param string|null $body Optional query body. For use with queries like INSERT with JsonEachRow format.
  * @return null|\stdClass|\stdClass[]|WP_Error
  */
-function clickhouse_query( string $query, string $body = '', string $return = 'auto' ) {
+function clickhouse_query( string $query, array $params = [], string $return = 'array', ?string $body = null ) {
```
rmccue, line 1103: I know the function's already here and named, but starting with a verb would be preferential imo (i.e. `query_clickhouse()`)

inc/api/namespace.php
roborourke: That was intentional, thought it might be useful.

inc/api/namespace.php
roborourke: Yeah there's a bunch I could abstract really but this pattern is only used in maybe 2 places. There's a bunch of other improvements in general to make anyway, like splitting some functions / queries up into multiple API endpoints etc..

inc/api/namespace.php
roborourke: Yeah, looking at some of the options that's probably best handled more intentionally via a filter if at all. I'll do this for now.

rmccue: If we switched to eg a native Clickhouse TCP client driver, I'd expect the call interface of this function to stay the same, but the params wouldn't need to be prefixed

inc/utils/namespace.php
roborourke: Realised the ES stuff was still in there so removed it and renamed back to just `Utils\query()`

inc/api/namespace.php
roborourke: Still uses the prefixes https://clickhouse.com/docs/en/interfaces/cli/#cli-queries-with-parameters

inc/audiences/namespace.php
```diff
@@ -490,31 +490,33 @@ function admin_enqueue_scripts() {
  * @return array|null
  */
 function get_estimate( array $audience ) : ?array {
-	global $wpdb;
 	$since = Utils\date_in_milliseconds( '-1 week', DAY_IN_SECONDS );
 
 	// Append the groups query.
 	$audience_where = build_audience_query( $audience );
```
rmccue, line 496: Feels like this should be returning a tuple of query and parameters?
```suggestion
	[ $audience_where, $audience_params ] = build_audience_query( $audience );
```

Then
```php
$query_params = [
	...$audience_params,
	'blog_id' => get_current_blog_id(),
```

inc/experiments/namespace.php
```diff
@@ -974,16 +976,28 @@ function process_post_ab_test_result( string $test_id, int $post_id ) {
...
 	// Process event filter.
-	$query_filter = '1=1';
-	if ( is_callable( $test['query_filter'] ) ) {
-		$query_filter = call_user_func_array( $test['query_filter'], [ $post_id ] );
+	$query_filter = $test['query_filter'] ?: '1=1';
+	$query_params = [
+		'blog_id' => get_current_blog_id(),
+		'ts' => $data['timestamp'] ?? 0,
+		'view_event' => $test['view'] ?? 'testView',
+		'goal_event' => $goal[0] ?? NULL_GOAL, // Special fallback value - if no goal is set the test will never convert.
```
rmccue, line 984: Add this comment to the constant too?

inc/utils/namespace.php
```diff
@@ -1145,10 +812,19 @@ function clickhouse_query( string $query, string $body = '', string $return = 'a
 		$request_args['body'] = $body;
 	}
 
+	// Add query parameters.
+	if ( ! empty( $params ) ) {
+		$param_keys = array_map( function ( $key ) {
+			return "param_{$key}";
+		}, array_keys( $params ) );
+		$params = array_combine( $param_keys, array_values( $params ) );
+		$clickhouse_url = add_query_arg( urlencode_deep( $params ), $clickhouse_url );
```
rmccue, line 821: I think this could be clearer with a foreach:

```suggestion
		$prepared = [];
		foreach ( $params as $key => $value ) {
			$prepared[ 'param_' . urlencode( $key ) ] = urlencode_deep( $value );
		}
		$clickhouse_url = add_query_arg( $prepared, $clickhouse_url );
```

`$params` should maybe also be json encoded or something?

inc/utils/namespace.php
```diff
@@ -1174,10 +850,21 @@ function clickhouse_query( string $query, string $body = '', string $return = 'a
...
+
+/**
+ * Get a unique cache key from a set of arguments.
+ *
+ * @param string $prefix The cache key prefix.
+ * @param mixed ...$args List of arguments to generate a cache key from.
+ * @return string
+ */
+function get_cache_key( string $prefix, ...$args ) : string {
+	return sprintf( '%s:%s', $prefix, hash( 'crc32', serialize( $args ) ) );
```
rmccue, line 869: Why crc32?

rmccue (changes requested):

inc/api/namespace.php
```diff
@@ -358,32 +358,37 @@ function get_graph_data( $start, $end, $resolution = '1 day', ?Filter $filter =
...
+	$query_params = [
+		'blog_id' => get_current_blog_id(),
+		'start' => (int) $start,
+		'end' => (int) $end,
+	];
 
 	if ( ! empty( $filter ) ) {
 		if ( $filter->type ) {
-			$types_where = array_map( function ( $type ) use ( $wpdb ) {
+			$types_where = array_map( function ( $type ) use ( $query_params ) {
```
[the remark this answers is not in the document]
roborourke: Ah, I thought pass by reference was implicit these days.

inc/api/namespace.php
```diff
@@ -207,20 +212,18 @@ function get_graph_data( $start, $end, $resolution = '1 day', ?Filter $filter =
...
+			$query_params['path_alt'] = $url_path_alt;
 		}
 	}
 
 	$query_where = sprintf( 'AND (%s)', implode( ') AND (', $query_where ) );
 
 	// Build SQL query.
-	$query = $wpdb->prepare(
+	$query =
 		"SELECT
```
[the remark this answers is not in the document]
rmccue: Ah, I didn't see any variables in there, maybe need new glasses

inc/api/namespace.php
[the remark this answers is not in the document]
rmccue: It's unclear whether you're mutating the variable itself here so not sure!

inc/audiences/namespace.php
roborourke: It's a bit of a PITA with the requirement for building some `ILIKE` queries so swerved it for now. I can revisit though.

rmccue: (Not clear from what the function is actually returning, but this is just a guess)

inc/utils/namespace.php
roborourke: Why json encoded? I'm not sure what you mean

rmccue: eg for a query like:
```php
Clickhouse\query(
    'SELECT * FROM analytics WHERE blog_id IN {blogs:Array(UInt16)}',
    [
         'blogs' => [ 1, 2, 3, 4 ],
    ]
)
```

Right now, you'll need to turn the array into `[1,2,3,4]` in your calling function; should `query()` handle this itself?

rmccue: (I'm not sure the format is actually JSON btw, so the answer is probably to make it the caller's problem)

inc/utils/namespace.php
rmccue: crc32 is a data integrity hash designed to ensure that your data hasn't been corrupted in transit/etc, and to my knowledge is not specifically designed to produce an [avalanche effect](https://en.wikipedia.org/wiki/Avalanche_effect); in other words, the likelihood of collision is much higher.

Performance isn't relevant for this few cycles, and you're right that security isn't really a factor here (specifically: reversibility of the hash); collision-avoidance is the most desirable property. Unless there's a good reason not to, I'd use `sha1()` which has good collision properties and is the most "standard" hash.

inc/utils/namespace.php
roborourke: Can we keep it simple for now? There are bigger fish to fry :)

inc/utils/namespace.php
roborourke: fair enough then, updating

inc/audiences/namespace.php
roborourke: Updated to return a tuple, string and params array.

inc/api/namespace.php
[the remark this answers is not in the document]
roborourke: I am, added anyway to be safe.

inc/utils/namespace.php
roborourke: Figured it out. It basically wants JSON encoding but with single quotes for strings e.g. `?param_foo=['a','b']`

roborourke:
Went through all the edits and got the array parameterisation working. Going to merge.

PR #514, opened 2022-09-23
Add broadcast post type and basic block

shadyvb:
Ref https://github.com/humanmade/product-dev/issues/1212


inc/blocks/broadcast/register.php
```diff
@@ -0,0 +1,110 @@
...
+			'window.Altis.Analytics.Broadcast.ManagerURL = "%s";',
+			'edit.php?post_type=broadcast'
+		),
+		'before'
+	);
+
+	wp_enqueue_style(
+		'altis-experiments-features-blocks-broadcast-style',
+		plugins_url( 'inc/blocks/broadcast/edit.css', Analytics\ROOT_FILE ),
+		[],
```
roborourke, line 71: is it dependent on the `wp-components` CSS?

inc/blocks/namespace.php
```diff
@@ -40,6 +41,10 @@ function setup() {
 		AB_Test_Variant\setup();
 		$using_blocks = true;
 	}
+	if ( Utils\is_feature_enabled( 'broadcast' ) ) {
+		Broadcast\setup();
+		$using_blocks = true;
```
roborourke, line 46: This variable should have perhaps been tailored to being about the requirement for the shim block - the shim block is only used for validating blocks that have conversion goals, e.g. checking for links / buttons etc... so I don't think you need it for the broadcast block.

inc/broadcast/namespace.php
```diff
@@ -0,0 +1,71 @@
+<?php
+namespace Altis\Analytics\Broadcast;
+
+use WP_Block_Editor_Context;
+use const Altis\Accelerate\PLUGIN_FILE;
```
roborourke, line 5: This puts a hard dependency on this plugin being part of the accelerate plugin. Can we avoid this?

package.json
```diff
@@ -75,5 +75,9 @@
     "webpack-dev-server": "^3.11.3",
     "webpack-manifest-plugin": "^2.2.0",
     "webpack-subresource-integrity": "^1.5.2"
+  },
+  "devDependencies": {
```
roborourke, line 79: This isn't a node project so these don't need to be dev dependencies.

roborourke (commented):
Looking good but I have a few questions and suggestions. One thing that does need to change is the dependency on the accelerate plugin_file constant.

inc/broadcast/namespace.php
shadyvb: Actually we don't use this anymore, so will remove.

shadyvb: We haven't really discussed where the menu item will be shown though, the icon is only needed if it'll be a top-level menu. Even then we can adjust the icon location to get rid of this const requirement.

inc/blocks/broadcast/edit.js
```diff
@@ -0,0 +1,230 @@
...
+		attributes,
+		setAttributes,
+	} = props;
+
+	const [ broadcast, _setBroadcast ] = useState( null );
+	const [ broadcasts, setBroadcasts ] = useState( [] );
+	const [ isFetching, setIsFetching ] = useState( false );
+	const [ searchKeyword, _setSearchKeyword ] = useState( '' );
+
+	const fetchBroadcasts = debouncePromise( function ( keyword, extraArgs ) {
```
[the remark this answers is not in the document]
shadyvb: I suppose not, was just copied from global-blocks before I moved everything within the analytics plugin. Will work on changing that 👍

inc/blocks/broadcast/edit.js
```diff
@@ -0,0 +1,230 @@
...
+			setSearchKeyword( '' );
+		}
+	}, [ broadcast ] );
+
+	return (
+		<Fragment>
+			<InspectorControls>
+				<div className="components-panel__body is-opened">
+					<Button
+						href={ window.Altis.Analytics.Broadcast.ManagerURL }
```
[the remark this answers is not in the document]
shadyvb: Added an icon for clarity.

inc/blocks/broadcast/edit.js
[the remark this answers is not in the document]
roborourke: Ah ok, just thinking they don’t actually need to be redeclared on render. Don’t worry about the WordPress data library stuff either for now, we can be pragmatic here

inc/blocks/broadcast/edit.js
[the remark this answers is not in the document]
roborourke: In accessibility terms an icon isn’t enough, think screen reader users. I believe if there are unsaved changes it will prompt you before navigating away anyway. Will have to see what it’s like to use I suppose

inc/broadcast/namespace.php
roborourke: What’s your feeling there? Next to global blocks?

Maybe we should have an Accelerate top level item with all features under it. Any thoughts on that @noeltock ?

inc/blocks/broadcast/edit.js
```diff
@@ -0,0 +1,228 @@
...
+		//  Input: " media "
+		keyword = keyword.trim();
+
+		_setSearchKeyword( keyword );
+	};
+
+	// Set the block clientId if none currently set.
+	useEffect( () => {
+		if ( ! attributes.clientId ) {
+			setAttributes( { clientId: uuid() } );
```
roborourke, line 101: Could look at the experience block example- blocks already have a UUID a level above the attributes so you can read from there instead of generating one. Not a blocker, just an alternative approach that means you don’t need the uuid library.

Also I learned recently browsers support this natively now: `window.crypto.randomUUID();`

https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID

inc/blocks/broadcast/edit.js
```diff
@@ -0,0 +1,228 @@
...
+		}
+	}, [ attributes.clientId, setAttributes ] );
+
+	// Fetch the broadcast post if we have one selected already.
+	useEffect( () => {
+		const abortController = new AbortController();
+		if ( attributes.broadcast && ! broadcast?.id ) {
+			fetchBroadcast( attributes.broadcast, { signal: abortController.signal } ).then( setBroadcast );
```
roborourke, line 109: Do we need a `catch` here too?

inc/blocks/broadcast/register.php
```diff
@@ -0,0 +1,112 @@
...
+ */
+function render_block( array $attributes, ?string $inner_content = '' ) : string {
+	$client_id = $attributes['clientId'] ?? null;
+	$broadcast_id = $attributes['broadcast'] ?? null;
+	$class_name = $attributes['className'] ?? '';
+	$align = $attributes['align'] ?? 'none';
+
+	// Warn if a client ID is not set.
+	if ( ! $client_id ) {
+		trigger_error( 'Broadcast content blocks must have a client ID set to track impressions. Please re-save the post content to generate one.', E_USER_WARNING );
```
roborourke, line 96: This raises an interesting question actually - we can track by broadcast ID and don’t technically need a clientId for the block, but that would give use data granularity to each block showing the broadcast.

im not sure how useful that is given we can group views by URL etc too. Fine to leave it in but I wouldn’t say it’s really required. And if it’s not set for some reason and we fill in a random one here we wouldn’t be able to use that data for analytics anyway.

roborourke (approved):
I think this is fine to get things moving though I had a few more thoughts looking through it again today. Up to you if feel they’re worth implementing, functionality wise though this good stuff

inc/blocks/broadcast/edit.js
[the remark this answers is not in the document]
shadyvb: I also added a title, but perhaps I forgot to say I also added a title.

inc/blocks/broadcast/edit.js
shadyvb: Makes sense, each block does have a client-id already. Will remove the attribute then.

shadyvb: Hmmm, seems like the clientId isn't persistent between sessions, so will keep the attribute but use the client id to fill it in instead of generating one.

inc/blocks/broadcast/register.php
shadyvb: Will leave it for now, pehaps we'll want to consolidate how tracking is done to use client-id as with other blocks. We'll discuss in more details in the tracking ticket.

inc/blocks/broadcast/edit.js
shadyvb: Added one and some handling for when the post is not available anymore. I let the user clear the selection and choose another one.

PR #530, opened 2022-10-25
Accelerate admin menu

shadyvb:
image

Ref https://github.com/humanmade/altis-accelerate/issues/100


inc/admin/namespace.php
```diff
@@ -56,3 +65,39 @@ function render_page() : void {
...
+ * @return void
+ */
+function register_admin_menu() : void {
+	add_menu_page(
+		__( 'Altis Accelerate', 'altis' ),
+		__( 'Accelerate', 'altis' ),
+		current_user_can( 'edit_posts' ),
+		'accelerate',
+		'Altis\\Analytics\\Dashboard\\load_dashboard',
+		plugins_url( '../assets/altis-logo.png', __DIR__ ),
```
roborourke, line 81: Can we get an SVG instead? What are dimensions of the PNG image?

roborourke (changes requested):
The image is too small / low res, remember high DPI displays etc... Code I think is good otherwise.

inc/admin/namespace.php
shadyvb: @roborourke Not sure how to generate a colored SVG from this, unless you mean using the plain colorless version?

Couldn't glean how to deal with high DPI displays from the WP admin css or dashicons, would just providing a bigger PNG and scaling it down with CSS work ?

roborourke: Found in drive here, was a little tucked away https://drive.google.com/file/d/1J3zrNuNLEk7P-US8xQGks9NJUMWqlOI4/view?usp=sharing

Re. high DPI displays yes, provide a larger image and scale down with CSS, effectively what `srcset / sizes` does.

roborourke (approved):

PR #531, opened 2022-10-27
Load images when Gaussholder is active.

robindevitt:
Related issue - https://github.com/humanmade/product-dev/issues/1246
Gaussholder functionality was added in - https://github.com/humanmade/Gaussholder/pull/71

This PR introduces the fix for when Gaussholder is active and the images don't load in Experience Blocks as well as when previewing a page.

#### Video showing the issue resolved

https://user-images.githubusercontent.com/16571365/198225048-2803259f-2166-4cca-b9e5-ac4c7fc8e711.mov

### DIFF
--- a/src/ab-test-preview.js
+++ b/src/ab-test-preview.js
@@ -55,6 +55,11 @@ function createTabbedPreviews() {
 
 				// append the data
 				tabContent.appendChild( variant );
+
+				// Ensure images are loaded with GaussHolder if available.
+				if ( window.GaussHolder ) {
+					window.GaussHolder();
+				}
 			} );
 
 			// append the created tab
--- a/src/ab-test-preview.js
+++ b/src/ab-test-preview.js
@@ -57,7 +57,12 @@ function createTabbedPreviews() {
 				tabContent.appendChild( variant );
 
 				// Dispatch the altisBlockContentChanged event.
-				window.dispatchEvent( new Event( 'altisBlockContentChanged' ) );
+				window.dispatchEvent( new CustomEvent( 'altisBlockContentChanged', {
+					detail: {
+						target: xb,
+						preview: true,
+					},
+				} ) );
 			} );
 
 			// append the created tab
--- a/src/ab-test-preview.js
+++ b/src/ab-test-preview.js
@@ -56,10 +56,8 @@ function createTabbedPreviews() {
 				// append the data
 				tabContent.appendChild( variant );
 
-				// Ensure images are loaded with GaussHolder if available.
-				if ( window.GaussHolder ) {
-					window.GaussHolder();
-				}
+				// Dispatch the altisBlockContentChanged event.
+				window.dispatchEvent( new Event( 'altisBlockContentChanged' ) );
 			} );
 
 			// append the created tab
--- a/src/audiences/preview/components/Selector.js
+++ b/src/audiences/preview/components/Selector.js
@@ -52,6 +52,11 @@ export default function Selector() {
 			setSelected( nextSelected );
 			Altis.Analytics.overrideAudiences( nextSelected );
 		}
+
+		// Ensure images are loaded with GaussHolder if available.
+		if ( window.GaussHolder ) {
+			window.GaussHolder();
+		}
 	};
 	/**
 	 * Prevent link default on click.
--- a/src/audiences/preview/components/Selector.js
+++ b/src/audiences/preview/components/Selector.js
@@ -52,11 +52,6 @@ export default function Selector() {
 			setSelected( nextSelected );
 			Altis.Analytics.overrideAudiences( nextSelected );
 		}
-
-		// Ensure images are loaded with GaussHolder if available.
-		if ( window.GaussHolder ) {
-			window.GaussHolder();
-		}
 	};
 	/**
 	 * Prevent link default on click.
--- a/src/experiments.js
+++ b/src/experiments.js
@@ -261,6 +261,9 @@ class ABTestBlock extends Test {
 		this.innerHTML = '';
 		this.appendChild( experience );
 
+		// Dispatch the altisblockcontentchanged event.
+		window.dispatchEvent( new Event( 'altisBlockContentChanged' ) );
+
 		// If variant ID is false then this viewer is not part of the test so don't log events.
 		if ( variantId === false ) {
 			return;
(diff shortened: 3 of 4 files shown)


shadyvb (approved):

shadyvb:
Also copied the same to Broadcast preview currently in the works 👍

shadyvb:
@robindevitt can you update the preview script here as well: https://github.com/humanmade/aws-analytics/blob/638c72f8beba7f28e55d07759aa698feff19020c/src/preview/altis-xb-preview.js#L62 , should be relatively the same as the `ab-test-preview.js` file.

robindevitt:
@shadyvb I have pushed up the suggestion 👍🏼 

I'm unsure if I should backport this too?

shadyvb:
@robindevitt I think the changes need to be backported, but not the one I suggested since it doesn't exist in previous versions yet, so might need to manually backport it to work.

src/audiences/preview/components/Selector.js
```diff
@@ -52,6 +52,11 @@ export default function Selector() {
 			setSelected( nextSelected );
 			Altis.Analytics.overrideAudiences( nextSelected );
 		}
+
+		// Ensure images are loaded with GaussHolder if available.
+		if ( window.GaussHolder ) {
+			window.GaussHolder();
+		}
```
roborourke, line 59: This code should be in the XBs themselves, triggered whenever the content is set / changed. You may still find race conditions without that specific coupling of the event order.

roborourke (changes requested):
I would take a different approach - GaussHolder may not be the only script someone may want to re-run when an XB changes its content. The problem can be solved here more generally rather than tying it specifically to GaussHolder.

We should dispatch a custom event instead on the window, similar to what you doing originally in the GaussHolder PR, e.g. `altisblockcontentchanged`.

Other code can then hook in e.g. 

```js
window.addEventListener( 'altisblockcontentchanged', function () {
    if ( window.GuassHolder ) {
        GaussHolder();
    }
} )
```

src/experiments.js
```diff
@@ -595,6 +595,9 @@ class PersonalizationBlock extends HTMLElement {
 				},
 			}, false );
 		} );
+
+		// Dispatch the altisBlockContentChanged event.
+		window.dispatchEvent( new Event( 'altisBlockContentChanged' ) );
```
roborourke, line 600: I think this is needed for the AB test block too?

robindevitt: Will have a look, this is currently in place for when previewing the page and toggling the tabs.

src/experiments.js
```diff
@@ -341,6 +341,9 @@ class ABTestBlock extends Test {
 				},
 			}, false );
 		} );
+
+		// Dispatch the altisblockcontentchanged event.
+		window.dispatchEvent( new Event( 'altisBlockContentChanged' ) );
```
roborourke, line 346: This is coming after quite a few return statements in the function, the content is changed further when `innerHTML` is set to empty and the template is appended as a child node. This means currently the block content could change but this event would not be dispatched.

src/experiments.js
roborourke: Still need to move this one :)

robindevitt: Sneaky! Got it moved accordingly now.

src/ab-test-preview.js
```diff
@@ -55,6 +55,9 @@ function createTabbedPreviews() {
 
 				// append the data
 				tabContent.appendChild( variant );
+
+				// Dispatch the altisBlockContentChanged event.
+				window.dispatchEvent( new Event( 'altisBlockContentChanged' ) );
```
roborourke, line 60:
```suggestion
				window.dispatchEvent( new CustomEvent( 'altisBlockContentChanged', {
					detail: {
						target: xb,
						preview: true,
					},
				} ) );
```

src/preview/altis-xb-preview.js
```diff
@@ -60,6 +60,9 @@ function createTabbedPreviews() {
 
 			// append the created tab
 			tabContainer.appendChild( tab );
+
+			// Dispatch the altisblockcontentchanged event.
+			window.dispatchEvent( new Event( 'altisBlockContentChanged' ) );
```
roborourke, line 65:
```suggestion
			window.dispatchEvent( new CustomEvent( 'altisBlockContentChanged', {
				detail: {
					target: xb,
					preview: true,
				},
			} ) );
```

src/experiments.js
```diff
@@ -261,6 +261,9 @@ class ABTestBlock extends Test {
 		this.innerHTML = '';
 		this.appendChild( experience );
 
+		// Dispatch the altisblockcontentchanged event.
+		window.dispatchEvent( new Event( 'altisBlockContentChanged' ) );
```
roborourke, line 265:
```suggestion
		window.dispatchEvent( new CustomEvent( 'altisBlockContentChanged', {
			detail: {
				target: this,
			},
		} ) );
```

src/experiments.js
```diff
@@ -530,6 +533,9 @@ class PersonalizationBlock extends HTMLElement {
 		this.innerHTML = '';
 		this.appendChild( experience );
 
+		// Dispatch the altisBlockContentChanged event.
+		window.dispatchEvent( new Event( 'altisBlockContentChanged' ) );
```
roborourke, line 537:
```suggestion
		window.dispatchEvent( new CustomEvent( 'altisBlockContentChanged', {
			detail: {
				target: this,
			},
		} ) );
```

roborourke (approved):
I was refreshing my memory just now on `Event()` and `CustomEvent()` and I believe it may be beneficial to provide some context with this event - this is a fine solution now and will do what we need but we can anticipate and support more fine grained integrations by providing some context with the events.

If we change to using `CustomEvent` and provide the current element as a `target` property people can scope their integrations per block if desired. I've added inline suggestions to apply, then we're good to merge.

Further reading on `CustomEvent` here https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent

roborourke:
@robindevitt thanks, good to merge and backport now.

The automated backport will fail but you can use it to get instructions for the manual process. If you make the manual backport PR against the latest release branch then you can add the additional backport labels to that manual PR. "Backporting the backport" essentially!

robindevitt:
@roborourke I added the labels and the backports have indeed failed and looking pretty messy to try and resolve. any tips of reverting to go with manual backport?

roborourke:
I'm not sure I follow? There's nothing to revert. Sometimes a backport isn't straightforward but it should be possible to work out, in this instance I think it's just one of the preview files that doesn't exist in the v4-6 branch? Or are there more issues than that?

If you start with a manual backport to the v4-6-branch then the resulting PR can be backported automatically rather than this one.

PR #532, opened 2022-10-27
Broadcast Manager UX improvements

shadyvb:
Ref https://github.com/humanmade/altis-accelerate/issues/102

### DIFF
--- a/src/accelerate/components/ListItem.tsx
+++ b/src/accelerate/components/ListItem.tsx
@@ -135,9 +135,12 @@ const ListItem = function ( props: {
 						</div>
 					</div>
 					<div className='record-name__title'>
-						<a href={ post.url || post.editUrl || '' } onClick={ (e: React.MouseEvent) => {
-							trackEvent( listId, 'Navigate', { type: post.type } );
-							onManage && onEdit( e );
+						<a href={ post.url || post.editUrl || '' } onClick={ ( e: React.MouseEvent ) => {
+							if ( onManage ) {
+								onEdit( e );
+							} else if ( post.url || post.editUrl ) {
+								trackEvent( listId, 'Navigate', { type: post.type } );
+							}
 						} }>
 							{ decodeEntities( post.title ) }
 						</a>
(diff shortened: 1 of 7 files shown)


src/accelerate/components/ListItem.tsx
```diff
@@ -131,7 +135,10 @@ const ListItem = function ( props: {
 						</div>
 					</div>
 					<div className='record-name__title'>
-						<a href={ post.url || post.editUrl || '' } onClick={ () => trackEvent( listId, 'Navigate', { type: post.type } ) }>
+						<a href={ post.url || post.editUrl || '' } onClick={ (e: React.MouseEvent) => {
+							trackEvent( listId, 'Navigate', { type: post.type } );
```
roborourke, line 139: This will trigger even if the href is empty and there's no navigation happening. I think it would be best to not have the anchor tag at all if there's no URL.

roborourke (changes requested):
One thing to change with regards to the list item title link, but otherwise code looks good 👍

src/accelerate/components/ListItem.tsx
```diff
@@ -131,7 +135,13 @@ const ListItem = function ( props: {
 						</div>
 					</div>
 					<div className='record-name__title'>
-						<a href={ post.url || post.editUrl || '' } onClick={ () => trackEvent( listId, 'Navigate', { type: post.type } ) }>
+						<a href={ post.url || post.editUrl || '' } onClick={ ( e: React.MouseEvent ) => {
+							if ( onManage ) {
+								onEdit( e );
```
roborourke, line 140: Given this change means there's no need to check for `onManage` in the `onEdit` function above it seems to me the entire `onEdit` function might as well be what's passed in for the `onManage` prop. That would remove code from this function that may never be used.

shadyvb: `onEdit()` is still used on links to edit content, even if `onManage()` is not defined/passed. And we still need to do the check either way because it's an optional callback.

roborourke: Aha, I missed the call to `onEdit` further down 👍

roborourke (approved):

PR #533, opened 2022-10-28
Sortable blocks within a Broadcast

shadyvb:
Ref https://github.com/humanmade/altis-accelerate/issues/101

### DIFF
--- a/inc/broadcast/namespace.php
+++ b/inc/broadcast/namespace.php
@@ -81,14 +81,13 @@ function register_rest_fields() : void {
 			$prev = array_map( 'absint', get_post_meta( $post->ID, 'blocks' ) ) ?: [];
 			$value = array_map( 'absint', $value );
 
-			$to_delete = array_diff( $prev, $value );
-			$to_add = array_filter( array_diff( $value, $prev ) );
-
-			foreach ( $to_delete as $block_id ) {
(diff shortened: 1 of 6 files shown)


inc/broadcast/namespace.php
```diff
@@ -81,14 +81,13 @@ function register_rest_fields() : void {
 			$prev = array_map( 'absint', get_post_meta( $post->ID, 'blocks' ) ) ?: [];
 			$value = array_map( 'absint', $value );
 
-			$to_delete = array_diff( $prev, $value );
-			$to_add = array_filter( array_diff( $value, $prev ) );
-
-			foreach ( $to_delete as $block_id ) {
-				delete_post_meta( $post->ID, 'blocks', $block_id );
+			if ( $prev == $value ) {
```
roborourke, line 84: Linter error here, will the strict equals affect the results of this check? Do the values need to be cast?:

```suggestion
			if ( $prev === $value ) {
```

roborourke (changes requested):

inc/broadcast/namespace.php
shadyvb: Nah, it's just a mistake.

roborourke (approved):

PR #534, opened 2022-11-01
Move analytics and insights to accelerate menu

shadyvb:
Ref https://github.com/humanmade/altis-accelerate/issues/98


roborourke (approved):
There's an annoying CSS linter error but we can just add an ignore rule comment for it.

shadyvb:
FYI this changeset has been copied to the great merge PR as well.

PR #546, opened 2022-12-13
Fix REST API schema output

roborourke:
Schema properties should have been callbacks.


shadyvb (changes requested):
LGTM, haven't checked if any other instances need to be fixed as well though. Got a tiny linting issue.

roborourke:
That's alright, it's on me to check and these were the only ones I found.

shadyvb (approved):

PR #556, opened 2023-07-11
Fix insights pages

roborourke:
Fixes issue with insights page not loading as results are semi-populated during API fetches.

Also resolves some issues with certain metrics not showing percentage symbols when they should.


mattheu (approved):
Approving. Looks OK... not that I have all that much understanding of what this is doing!

roborourke:
the code would break on `test?.results?.variants[ variant.id ]` when `test.results = {}`, it would try to access a non existent array. Was handled further up in the file the same way I changed it to. Not sure how it got missed at the time, been broken for ages!

